PR #610, opened 2016-12-06
Ticket - 18457 Add column to OU1 & OQ1 to assist in order quantity tracking (On Hand Not Allocated)

sewa67:
Files - oe/b-ordinq.w,oeinq/b-ordinq.w

### DIFF
--- a/Legacy/oe/b-ordinq.w
+++ b/Legacy/oe/b-ordinq.w
@@ -81,6 +81,7 @@ DEF VAR lc-mi AS CHAR NO-UNDO.
 DEF VAR lr-rel-lib AS HANDLE NO-UNDO.
 DEFINE VARIABLE dTotQtyRet AS DECIMAL NO-UNDO.
 DEFINE VARIABLE dTotRetInv AS DECIMAL NO-UNDO.
+DEFINE VARIABLE iHandQtyNoalloc AS INTEGER NO-UNDO .
 DEF VAR lActive AS LOG NO-UNDO.
 DO TRANSACTION:
      {sys/ref/CustList.i NEW}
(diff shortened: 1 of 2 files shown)


wade-advantzware (changes requested):
Sewa, the formatting guidelines call for functions to begin with 'f':
Procedure blocks begin and end should align – use Source>Correct Indentation in Progress Developer Studio 
Place a comment on all END. statements indicating which procedure block it belongs to and that it is indented to match the starting statement  
Use properly named variables that are descriptive  
All Progress key words are capitalized – Run Source>Correct Case in Progress Developer Studio 
Expand all Progress Key words – Run Source>Expand Keywords in Progress Developer Studio 
Either EXCLUSIVE-LOCK or NO-LOCK appears on the same line as the table is accessed, ie, FOR EACH <TableName> NO-LOCK  
Lock status should appear before the WHERE clause and not at the end of the statement  
All local variables should begin with its data type “c Character, d Decimal, dt Date, i Integer, l Logical, r RowID, h Handle”, etc.  
Use camel case naming conventions, ie, a decimal variable for Price Amount = dPriceAmount  
Internal procedures start with a lower case “p<ProcedureName>”  
Functions begin with “f<FunctionName>”  
Temp-Tables begin with “tt<TempTable>”  
INPUT PARAMETER for a CHARACTER would be ipc<VariableName>  
OUTPUT PARAMETER would be opc<VariableName>  
INPUT-OUTPUT PARAMETER would be iopc<VariableName> 
The file must open in AppBuilder without errors! 
Avoid use of ROWID and do not use RECID.  Never store a ROWID value for use later, unless it is a very short time period.

PR #611, opened 2016-12-06
Ticket - 13487 AQ1 Excel Export - Extremely Slow

sewa67:
Files - arinq/rd-invexp.w


wade-advantzware (changes requested):
Sewa, per the guidelines, the no-lock should be right after the 'FIRST ar-inv', and the period ending the assign statement should be on it's own line.

PR #1205, opened 2017-06-26
Feature/21045 implementation of edi graphic packing and lutron

wade-advantzware:


tyndmar (approved):
As far as I can tell, this is clean code.  Not qualified to determine if it meets requirements.  Will advise of any compilation issues after merge.

PR #2024, opened 2018-01-29
Ticket - 23907 NK1 OEPriceMatrixCheck - Setting to *block* the addingof the order at a price/quantity not on the matrix

sewa67:
Files - oe/GetPriceMatrix.p,oe/d-oeitem.w,oe/oe-pric2.i,sys/ref/sys-ctrl.i


wade-advantzware (changes requested):
eff-date filter should be retained

PR #2092, opened 2018-02-15
#25494 Reftable - [dropslit]

tyndmar:


brad-advantzware (commented):
est.dropslit = reftable.loc eq "yes" misplaced and assignment into est.recalc also looks like it was removed incorrectly

PR #2094, opened 2018-02-15
#25503 Reftable - oe-rel.s-code

tyndmar:


brad-advantzware (commented):
No conversion program?
oe/actrelmerg.p reft-s-code still being referenced - is this a partial conversion and the reft-s-code is not changed yet for another reftable?
Otherwise, looks pretty good.  THis should be a big code simplifier and maybe performance improver

PR #2095, opened 2018-02-15
#25504 Reftable - [oe-relh.can-print]

tyndmar:


brad-advantzware (commented):
OK.  Considering the repetitive definition of temp-table, this would be better in a .i

PR #2206, opened 2018-03-03
Ticket - 26172 Add Job Suffix to M-U-6

sewa67:
Files - rm/rmte&p.w


brad-advantzware (changes requested):
Why is the concatenation of the job-no2 into job-no required for this?  Why not just use the job-no2 fields directly with the 2 integer fill ins?

PR #2214, opened 2018-03-05
26546 Cannot save menu buttons after making changes - error message

ron-advantzware:


brad-advantzware (commented):
Ron - I don't understand this change,  1275 lines removed from usermenu.w?

PR #2215, opened 2018-03-05
26736 RFID Monitor

ron-advantzware:


brad-advantzware (approved):
Initializing before import as indicated in Ron's ticket comment

PR #2222, opened 2018-03-06
Ticket - 26842 Issues with Order Entry

sewa67:
Files - oe/d-oeitem.w


brad-advantzware (changes requested):
I do not understand what circumstances all line items for an order should be deleted.  Need to justify this (in ticket)

brad-advantzware (approved):

PR #2223, opened 2018-03-06
Ticket - 26844 When you delete a location, the screen does not refresh so the location still appears

sewa67:
Files - viewers/usrcmp-l.w


brad-advantzware (changes requested):
Mark - I generally don't think manipulating adm global variables is a good practice, if you agree, please fail this for Sewa and suggest an alternative

PR #2224, opened 2018-03-06
Ticket - 26764 Error when Scan Job # for Set Component into Loadtag creation

sewa67:
Files - oerep/r-loadtg.w


brad-advantzware (approved):
Looks like this passes an indicator to search for blank into loadtag processing

PR #2225, opened 2018-03-06
Ticket - 25536 Request - Default Receiving Location.

sewa67:
Files - viewers/itemfg.w


brad-advantzware (approved):
This change includes a "blank" as an invalid warehouse which is good, even if, for some reason, there is a blank location code in the system

PR #2285, opened 2018-03-10
Ticket - 26691 Populate the itemfg.setupDate field when FG item is created and make sure utility does not inactivate new items.

sewa67:
Files - util/setfgia.w, viewers/itemfg.w


brad-advantzware (changes requested):
This appears to only be set when FG Items are created through the Viewer UI.  Need to also assign this value when items are created through Order/Job Entry (create an order or job with estimate that has a blank FG Item#)

PR #2302, opened 2018-03-13
Ticket -26834 Business Form Capability - Ask asi user for format on execution

sewa67:
Files - oerep/r-acknow.w,windows/l-syschrL.w


brad-advantzware (changes requested):
Finding the ASI user really means using the existing procedure to confirm that the user is security level >=1000.
Use new ProgramMasterSecurity.p program

PR #2310, opened 2018-03-14
#24790 Reftable - FLUTE

jgill-advantzware:


brad-advantzware (commented):
Mark - Is the removal of these Open-Query blocks going to impact things?  They seem to be already commented

PR #2357, opened 2018-03-28
27456 EDI edivtran.w incorrect external table

ron-advantzware:


brad-advantzware (approved):
Confirmed that size adjustments were intentional

PR #2367, opened 2018-03-30
Ticket - 27375 AQ1 Does Not Complete

sewa67:
Files - arinq/rd-invexp.w


brad-advantzware (approved):
I guess this is acceptable, although I'm not sure why converting from a dynamic field display to discrete cases will matter.  Seems like Sewa's primary fix is lLineTotal logical, that will not process line specific data if those columns aren't selected

PR #2376, opened 2018-04-03
Ticket - 26278 Set estimate - Prompt Warning if the same customer part # is on multiple items in a set estimate

sewa67:
Files - cec/d-updset.w, est/crt-set.w


brad-advantzware (approved):
Approved since Sewa did as asked.  I questioned spec in ticket

PR #2388, opened 2018-04-04
Ticket - 27605 Redundant Message on OVERRIDE.

sewa67:
Files - cec/v-est.w


brad-advantzware (approved):
There are other issues with the METRIC flag in estimating.  Commit this so that this can be taken into account with my work in that area

PR #2430, opened 2018-04-10
27821 Posting RM - Error messages when entering a single digit for Jo…

ron-advantzware:
…b2 fields


brad-advantzware (commented):
I don't see changes to the use of the variables that were changed from Char to Int.  Was that in a previous check-in or not necessary?  I would assume seeing some INT() function wrapping the old variable, that should no longer be necessary

PR #2439, opened 2018-04-11
27850_EDI_-Order_Line_Numbers_Misaligned

wade-advantzware:


brad-advantzware (changes requested):
Wade - after our conversation - I think the assignment of the counter from cust-po-line should be modified so as to use the "standard" order line, propagated through the OE process of pulled from oe-ordl if necessary

PR #2443, opened 2018-04-12
27821 Posting RM Validate end_job-no2 Field

ron-advantzware:


brad-advantzware (changes requested):
Do the same for the begin job 2

PR #2465, opened 2018-04-16
27487 Modifications to Amazon EDI options

wade-advantzware:


brad-advantzware (changes requested):
Dev messages in Local-disable

brad-advantzware (changes requested):
Still a developer message in local-disable  ("disable it").  Must be more careful

PR #2466, opened 2018-04-17
Ticket - 28063 AU1 - A few more minor changes to be made

sewa67:
Files - ar/d-arinvl.w,ar/v-arinv.w


brad-advantzware (changes requested):
Sewa - I will have to defer to the tester regarding the real-time calculation change.  Need changes to the code to set the default due-date.  First, the naming convention of the variable is very misleading.  oeDateAuto is A COMPLETELY DIFFERENT NK1.  Rename that variable to reflect the InvAddDate nk1.  Also, the logic of the due-date should be based on the terms.  For a simple outline of logic, refer to the "pCreateNewInvoiceAR" proc in util/ImportAR for logic to find the appropriate terms code (use terms.net-days).

PR #2469, opened 2018-04-17
Ticket - 27979 I-C-2 Copy function.

sewa67:
Files - fg/b-phys.w


brad-advantzware (commented):
Will require testing to confirm but changing to Default Action (double-click) now triggers "new-state"

PR #2472, opened 2018-04-17
#27539 Deleting Zero Bins NF10

tyndmar:


brad-advantzware (commented):
Mark - Please schedule a review of this with me.  I'm not clear on the changes you made and they seem significant

PR #2499, opened 2018-04-20
Ticket - 28108 M-F-1 History Tab Double Click to Change History should not Demand PO#

sewa67:

Files - rminq/d-rmiinq.w


brad-advantzware (changes requested):
Sewa - The first validation is included in the second so it can be removed.  If the PO is entered, per the ticket, it needs to be valid for the item.

PR #2502, opened 2018-04-20
Ticket - 25917 Need Invoice Print format that lists PO by Line Item.

sewa67:
Files -  ar/rep/invxprnt10.i, ar/rep/invxprnt10.p, oe/rep/invxprnt10.i, oe/rep/invxprnt10.p


brad-advantzware (changes requested):
Changes / spec under review by PM (Clay)

brad-advantzware (commented):
Approved but I have notified Clay to examine this more closely.  I don't think the invoice should need to go back to the release to get the PO.  The PO is already on the invoice line (and is editable).  Move this to Testing

PR #2509, opened 2018-04-20
28147 Farm Out Tab in EC

ron-advantzware:


brad-advantzware (changes requested):
Dev message in v-eitem2.w

PR #2511, opened 2018-04-20
26031 job xml monitor changes

wade-advantzware:


brad-advantzware (changes requested):
Assigned to Ron for Code Review of resize code

PR #2519, opened 2018-04-24
27258 AOA Audit History Report

ron-advantzware:


brad-advantzware (commented):
Same feedback as the recent block of code copied into monitor to do the window resize.  Is there not a better way to create a central procedure for things like user-print, or is it completely unique to this application

PR #2525, opened 2018-04-24
Ticket - 25536 Request - Default Receiving Location

sewa67:
Files - viewers/itemfg.w


brad-advantzware (changes requested):
Default-loc procedure should not be necessary.  Move the "Customer X" shipto logic in the create trg to

PR #2526, opened 2018-04-24
Ticket - 27735 N-K-1 RMUnderOver -- M-U-1 Receipt Overrun Calculation is Using View Tab of P-U-1

sewa67:
Files - rm/b-rcptd.w


brad-advantzware (changes requested):
Sewa - Is validate-qty-over copied and pasted from chkporun.i?

brad-advantzware (changes requested):
Spoke to Mark regarding the use of copy and pasted code

PR #2535, opened 2018-04-26
28124 Saving changes to copied matrices items on the Vend Cost tab of…

ron-advantzware:
… IF1


brad-advantzware (approved):
This seems like it will work, but seems like a non-standard CRUD.  We may be too far gone on this to get back to your framework, but I assume this a departure from that?

PR #2540, opened 2018-04-26
Ticket - 28192 error message when adding AP invoice with a future invoice date

sewa67:
files - ap/v-apinv.w,apinq/b-apinq2.w


brad-advantzware (approved):
Sewa - ok, but please confirm that there is only one call to record-added that has to have this new parameter

PR #2548, opened 2018-04-27
Ticket - 25917 Need Invoice Print format that lists PO by Line Item.

sewa67:
Files - ar/rep/invxprnt10.p, oe/rep/invxprnt10.p


brad-advantzware (changes requested):
Mark - This may be a simple revert.  I believe Sewa did not receive the correct spec on this, so we may want to reject.  Clay will update in ticket

PR #2550, opened 2018-04-27
Ticket - 26307 Request - DSG BOLs

sewa67:
Files - oe/rep/bolprem.i,oe/rep/bolprem23.i,oe/rep/bolprem3.i,sys/inc/bolform.i


brad-advantzware (approved):
these are xprint forms, so I'm not clear why we are manipulating line counts.  We should be controlling paging through Xprint.  But this doesn't make anything worse.

PR #2571, opened 2018-04-30
28373 AU1 Invoice Qty does not appear on printout

wade-advantzware:
1) Set ship-qty to print on invoices
2) Corrections to file so it would load into app builder


brad-advantzware (commented):
Wade - Assigning ar-invl.ship-qty from ar-invl.inv-qty is problematic.  In an invoice-only, there are no goods shipped.  Is this only on newly added invoices where the assumption is that what is invoiced was "shipped".  Is there not a way to allow a user to change this value?

PR #2573, opened 2018-05-01
Ticket - 28368 Job Creation Date.

sewa67:
Files - jc/v-job.w


brad-advantzware (changes requested):
Create Date assignment should be in create.trg/job.p

brad-advantzware (changes requested):
Create-date assignment added to trigger but not removed from viewer.  Remove redundant assignment from viewer

brad-advantzware (commented):
Do not need Find Current stuff - needs to be rejected

PR #2574, opened 2018-05-01
Ticket - 28079 FG Receipt Error on Set Headers.

sewa67:
Files - fg/b-rcptd.w


brad-advantzware (changes requested):
NO-ERROR prior to DELETE fg-rctd will not prevent error being fired on Delete of record if it is not available

brad-advantzware (approved):

PR #2578, opened 2018-05-02
Ticket - 28405 Sharp Shooter Cycle Count / Bin Location Changes the Cost UOM

sewa67:
Files - addon/fg/b-phys.w


brad-advantzware (changes requested):
Sewa - The problem lies within the fgrctdcst.p program.  At the end, there is a condition that sets the UOM to "EA".  The fall back should be to set the UOM to the FG Item, Cost UOM (bottom right of IF1 View-Screen)

brad-advantzware (approved):

PR #2579, opened 2018-05-02
28443 EDI 810 for Amazon should not need to have partner AMAZ set up

wade-advantzware:


brad-advantzware (changes requested):
Why is emast Exclusive-lock with the OF condition and No-lock, with the match on ws_partner?

brad-advantzware (approved):

PR #2583, opened 2018-05-03
Ticket - 28301 Invoices off center.

sewa67:
Files - oerep/r-bolprt.w


brad-advantzware (changes requested):
Sewa Singh Clarification:
If a business form already has a "LEFT" override, add the decimal value of the NK1 to the mm measurement (could be STRING(7 + Decimal-val), (6 + Decimal-val).
If a business form does not already have a LEFT override, only apply the LEFT override if decimal is >0 and apply the left override fully:  STRING(Decimal-val) without a hard-coded offset.  Do not even use the LEFT tag if the decimal value is 0

brad-advantzware (changes requested):
Almost - The standard format condition looks correct.  For those formats "Carded", "CCC", etc.  you should add the decimal value to the value that is already there, which will allow them to tweak the margin as a difference from what they already hard-coded.

brad-advantzware (approved):

PR #2598, opened 2018-05-08
Ticket - 28448 Creating Loadtags mode, with auto close checked does not clear user from user table on exit

sewa67:
Files - oerep/r-loadtg.w


brad-advantzware (changes requested):
Use userLogOut.p to remove session (based on session) ID.

brad-advantzware (approved):
Not ideal but acceptable.  Already discussed other strategies for this and since we can't do OO, I guess this is fine

PR #2600, opened 2018-05-08
#25485 Reftable - blank-vend-qty

jgill-advantzware:


brad-advantzware (changes requested):
Jitender -need to look at this one closer.  The e-item.run-cost is already in user and being assigned to the corresponding index.  The reftable was used to add 10 more array elements.  Conversion would overwrite the existing 10 e-item.run- records and that will not work.  Because e-itemf.run-qty and .run-cost have max extent 10, this will need to be a DB change to create a second set of variables to store the 11-20 values.  Please review again and confirm if you agree

PR #2615, opened 2018-05-10
Ticket - 28699 Job Ticket Prints All Spec Notes regardless of Paramter Setting.

sewa67:
Files - cecrep/jobtick.p, cecrep/jobtickL.p


brad-advantzware (approved):
There are 2 fillins on main job screen:  one for spec codes and one for departments. This one appears to only use the dept-codes list.  There is also a shared var for "spec-list"  I'll approve this change but alert testing to the fact that this only changes the sensitivity to the dept-list selection (which is needed anyway).  Expect to fail testing.

PR #2619, opened 2018-05-10
28731 TEST - Hot Keys

ron-advantzware:


brad-advantzware (approved):
Only change I see is addition of RETURN NO-APPLY.  Other changes cosmetic

PR #2623, opened 2018-05-11
28702 Two Blank Vendor Farm Tabs when adding FG during Estimate Entry

ron-advantzware:


brad-advantzware (approved):
Ron - Does this prevent the problem of the initial record?  Or does the "3rd" record not get created now because it can match the .i-no?  Approved so that we can test

PR #2625, opened 2018-05-11
Ticket - 28601 OS7 need to export completely to Excel.

sewa67:
Files - oerep/trkschxl.p, viewers/carrier.w


brad-advantzware (changes requested):
Sewa - Why do we need to introduce a new .xlt?  If this is a defect, the existing template should just be fixed

PR #2628, opened 2018-05-11
Ticket - 28210 EF / Flat Packing Hours added to the job card (Machinespeed, quantities, totaled like other items on card)

sewa67:
Files - cerep/jobpryst.p,resources/template/PrystupJob.xlt


brad-advantzware (changes requested):
The Gluer Speed assignment from the job should not be removed.  Also, if we moved the location of the data, all Excel references after that move will have to be changed.  Unable to review template through GitHub but this should've just added the data entry on the last empty spot rather than moving others down.

brad-advantzware (approved):
Just get it to a tester

PR #2636, opened 2018-05-12
Ticket - 28427 Ensure consistent Vend Cost decimal value from FG Item | Vend Cost to Order to auto PO

sewa67:
Files - po/d-vndcst.w,po/dopo.p


brad-advantzware (changes requested):
Sewa - Why was the AND AVAILABLE e-item-vend removed?
Also, (this isn't your change) but this block makes no sense:
IF AVAILABLE e-item-vend THEN
           cUom = e-item-vend.std-uom.
       ELSE IF AVAILABLE e-itemfg-vend THEN
           cUom = e-itemfg-vend.std-uom.

brad-advantzware (approved):
Sewa explained change in ticket

PR #2643, opened 2018-05-14
#29632 Reftable - e-itemfg-vend.std-uom

jgill-advantzware:


brad-advantzware (changes requested):
Need to review with Jitender.

brad-advantzware (changes requested):
Legacy/sys/inc/fgcascnt.i  make sure you assign the e-itemfg-vend.std-uom to e-itemfg.

brad-advantzware (approved):
reftableMigration should be a for each through all matching e-itemfg-vend tables

brad-advantzware (approved):

PR #2644, opened 2018-05-14
28772 Do not allow any user other than ASI to change the system date

ron-advantzware:


brad-advantzware (approved):
Disabled per ticket (not active for ASI only as in subject)

PR #2645, opened 2018-05-14
28247 Vorne barcode won't print on the job cards

ron-advantzware:


brad-advantzware (approved):
I remember hard-coding that because I couldn't get it to work otherwise for some reason.  Knew that would come back to be a problem

PR #2646, opened 2018-05-15
28666 VU3 AOA Post

ron-advantzware:


brad-advantzware (changes requested):
I think you targeted the right code.  Just need to also update the "classic" post program too.

brad-advantzware (approved):

PR #2650, opened 2018-05-15
Ticket - 20657 Release Folder Import from Excel File.

sewa67:
files - util/Imprel.p, util/ImportRelease.p, util/ttImport.i

### DIFF
--- a/Legacy/util/ImpRel.p
+++ b/Legacy/util/ImpRel.p
@@ -1,0 +1,1 @@
+{util/ImpInclude.i ""impRel."" ""ttImportRelease""}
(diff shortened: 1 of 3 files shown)


brad-advantzware (changes requested):
Variable names don't conform to standards and you are supposed to run the OE architect formatting of "Expand Keywords", "Correct Case" and "Correct Indentation".
Example variable definition:
DEF VAR v-qty-sym as int no-undo.
should be:
DEFINE VARIABLE iQtySum AS INTEGER NO-UNDO.
Buffers should also include the full table name - bf-rel -> bf-oe-rel.
You also switch between using EQ and = in conditions.  Use EQ.

PR #2651, opened 2018-05-15
Ticket - 29031 Routing Matrix Problem

sewa67:
Files - est/vstyrout.w


brad-advantzware (changes requested):
I don't know what is being attempted here.  Adding comment for Sewa and Jay in ticket

brad-advantzware (changes requested):
I don't want the initialize code in there that auto-numbers the MSF.  Instead, check for one with an MSF = 0 before add is dispatched and required the user to change that MSF.  "A routing matrix with 0 MSF exists.  Please change the MSF for that to something other than 0."  Remove the string concatenation with entry code block

brad-advantzware (approved):

brad-advantzware (approved):

PR #2653, opened 2018-05-15
28831 Computer DC locks up after 12 am

ron-advantzware:


brad-advantzware (commented):
Do we not already have the TSBreaks logical in this program?  If not, we are trying to encapsulate the sys-ctrl queries into the nk1look.p program (for lack of a better place).  I would prefer no direct query to sys-ctrl.  I know there are already a lot out there, but the fewer the better.

PR #2654, opened 2018-05-15
#25449 Reftable - e-itemfg-vend.markup

jgill-advantzware:


brad-advantzware (changes requested):
This looks correct based on current code, but knowing what I now know about the e-itemfg-vend table, there is a ton of ambiguity in this.  There may be (likely) multiple e-itemfg-vend records that match for each company, vend, and i-no.  We need to target these as .est-no eq "" so that it picks up the Vend Cost tab and not the Farm Tab matrices.

PR #2662, opened 2018-05-16
28373_AU1_Qty_not_on_printout_add_ship_qty

wade-advantzware:
Could not open d-arinvl.w as it was so had to open it and save it to fix the issues.


brad-advantzware (commented):
Wade - too difficult to review.  can you summarize or point me to the main changes?

brad-advantzware (approved):
Too many UI related changes to review code. Approving to pass on to testers

PR #2663, opened 2018-05-17
Ticket - 28495 IF1 Vend Cost Tab to have Import Option.

sewa67:
Files - util/ttImport.i, util/ImportVendorCost.p, util/ImpVenCst.p


brad-advantzware (changes requested):
Sewa - I think this looks OK.  At least one change right now:  Instead of reassigning all variables in an Assign statement, can you do a BUFFER-COPY of the one already found/created?  I didn't compare line by line but there appears to be a lot of copied code here.

brad-advantzware (changes requested):
do not increment iopiAdded for the "Farm Tab" Create record, just do it once at the top of pProcessRecord

brad-advantzware (approved):

PR #2667, opened 2018-05-19
Ticket - 28382 Should we Reverse B and S

sewa67:
Files - ce/v-est2.w,cec/v-est2.w


brad-advantzware (changes requested):
Sewa - see specs in ticket, "Sheet" should store B, "Blank" should store "S"

brad-advantzware (approved):
Approved (but have concerns to take up with Prod Dev)

PR #2673, opened 2018-05-21
Ticket - 29032 Glue Joint not updating on Box Design Tab if the TAB key is not Pressed after changing the Joint Tab Width.

sewa67:
Files - cec/v-est.w


brad-advantzware (approved):
Odd variable name but it is consistent with the ON LEAVE

PR #2675, opened 2018-05-21
#25449 Reftable -e-itemfg-vend.markup

jgill-advantzware:


brad-advantzware (changes requested):
See notes inline in code (DO i = 1 to 10 and PARAM BUFFER)

brad-advantzware (changes requested):
Do not name the buffer io-  name it bf- if not a param buffer

brad-advantzware (approved):

PR #2678, opened 2018-05-21
28570_Store_entered_data_to_database_in_create_BOL

wade-advantzware:


brad-advantzware (commented):
Wade - what is the major change here?  It looks like the program was already saving ssrelbol to db.  Is this to do it more reliably/frequently?  Is the use of the temp-table necessary at all?

brad-advantzware (approved):
Approving for testing.  Wade - please set up some time to review this with me tomorrow

PR #2682, opened 2018-05-22
Ticket -29306 Change Price Matrix Import to new Import Infrastructureand include Ship To ID in import columns.

sewa67:
Files - util/ttimport.i, util/ImportPriceMatrix.p, util/ImpPMtx.p


brad-advantzware (changes requested):
ExpDate and ShipTo are last on the report, and I see that is matched in Export file.
Only change is to update comments and header info

brad-advantzware (changes requested):
Additionally, we need to replace the current Import program with this one (OF3)

brad-advantzware (approved):

PR #2685, opened 2018-05-22
29302 Unable to Enter a Space in OT1 Customer Number

ron-advantzware:


brad-advantzware (approved):
I assume this works but why do have any code that tests the key value entered? What is the purpose of the cursor-offset line?

PR #2686, opened 2018-05-22
#29056 - Too long to return error on bad password

tyndmar:


brad-advantzware (commented):
What is this change doing?  Does it not "retry" to connect again?  Is it a problem that the connect does not have NO-ERROR in this iteration?

brad-advantzware (approved):
OK, assuming NO-ERROR not required on CONNECT statement and that error is handled in message in caller when RETURN is invoked.

PR #2687, opened 2018-05-22
Feature/25583 po routines to check that folder exists

wade-advantzware:


brad-advantzware (changes requested):
Wade - Can this folder be an NK1? ".\custFiles\EDIFiles\POs\"  In general, I think we should avoid any hardcoded path for non program files.

brad-advantzware (approved):

PR #2688, opened 2018-05-22
28373_AU1_Qty_not_on_printout_add_ship_qty

wade-advantzware:


brad-advantzware (approved):
Difficult to tell what the actual change it but it appears to be just adding fields to ui

PR #2690, opened 2018-05-22
#27968 Problem with "admin" user

tyndmar:


brad-advantzware (changes requested):
Mark - Similar code should be added to user groups as well.  Anywhere asi appears should also add admin.  Raising a question for product dev

brad-advantzware (approved):

PR #2694, opened 2018-05-23
28386 EDI Checkbox enabled only if EDI module enabled

wade-advantzware:


brad-advantzware (changes requested):
Wade - Do not create a new ModuleACtive.p file.  Please review CheckModule.p which should do the same.

brad-advantzware (changes requested):
Remove moduleActive from source

brad-advantzware (approved):

PR #2695, opened 2018-05-23
Ticket - 25963 Pricing at Order Entry

sewa67:
Please review  this change and merge into development branch 
Files - oe/d-oeitem.w,oe/v-ord.w


brad-advantzware (changes requested):
Cannot assume that oe-ord.ship-id is not blank.  Need to include a condition that only uses the oe-ord.ship-id if it is not blank

brad-advantzware (approved):

PR #2698, opened 2018-05-23
Ticket - 29038 Status Codes - Documentation

sewa67:
Files - oe/b-ordinq.w,oe/vi-ord.w


brad-advantzware (changes requested):
Code to get the status should not be copied into 2 places.  Create new program GetStatusDesc.p  
Use Case or lookup instead of sequences of Ifs

brad-advantzware (approved):

PR #2701, opened 2018-05-23
29323 AOA Invoice Posting not running

wade-advantzware:


brad-advantzware (commented):
Wade - Need to review with you.  2 things I don't like, in general - shared variables and hard-coded file paths

brad-advantzware (approved):
Discussed with Wade.  Global Shared variables need to be created.  We should avoid this in the future, but this is how the EDI code was done originally

PR #2702, opened 2018-05-24
29358 AOA Reports using field rowType in index exceed index size limit

ron-advantzware:


brad-advantzware (approved):
Might this result in a slower report run?  Is rowType used in all AOA reports?

PR #2705, opened 2018-05-24
Ticket - 28301 Invoices off center.

sewa67:
Files - oerep/r-invprtProc.i


brad-advantzware (approved):
OK for now.  Next time this is edited, we should look to simplify the xprint prefix code.  Each tag should have it's own condition/logic and concatenated into a single header string rather than a big string of if/then/elses

PR #2723, opened 2018-05-25
Ticket - 28341 N-K-1 JobCardF = Wingate Need to Print Lineal Footage on Job Ticket

sewa67:

Files - cerep/jobwin2.p


brad-advantzware (changes requested):
mach feed type needs to be logic that triggers LinearFeet = YES.  Current change uses item.mat-type which is "REAM" and is not correct

brad-advantzware (approved):

PR #2728, opened 2018-05-25
29311_Add_security_for_Price_in_Order_Item_and_Price_Hold_on_Header

wade-advantzware:


brad-advantzware (approved):
So the code was doing a direct - Apply Entry?  Should the OESCREEN option really just change the tab order rather than trying to manage what occurs on-leave?  Don't need to fix that now, but curious about a better way to do this

PR #2733, opened 2018-05-28
Ticket - 29556 Cost should be calculated the correct way - Look for invoice line item cost and if not there then order line cost.

sewa67:
Files - oeinq/b-ordinq.w


brad-advantzware (changes requested):
Sewa - Existing cost (ld-cost) column should be changed to "Invoice Line Cost"

brad-advantzware (approved):

PR #2736, opened 2018-05-29
Ticket - 26485 GL# when adding PO for FG item should import from I-F-3 with Job# does not exist

sewa67:
Files - po/d-poordl.w


brad-advantzware (commented):
Hold on merge - Requested full logic breakdown from Sewa

brad-advantzware (approved):

PR #2738, opened 2018-05-29
Ticket - 29320 Quotes for Inactive Items should not print using EQ.

sewa67:
Files - est/b-qprice.w, est/b-qtitm.w, fg/d-invprc.w


brad-advantzware (changes requested):
This is still In Process.  I don't see changes to Quote Printing (new parameter - Active/Inactive)
Nor do I see filtering of Active-only items.  This looks like it is just adding columns and warnings

brad-advantzware (changes requested):
Sewa - quoxptree.i - Directly joining itemfg will require the quote to have a finished good.  This may not be the case.  Change should be to go back to the find first itemfg, and test the stat.  If avail and stat eq "I" then NEXT.

brad-advantzware (approved):

PR #2752, opened 2018-05-31
Ticket - 29628 NK1 RMUnderOver Logical Value should Turn this Feature On or Off.

sewa67:
Files - rm/chkporun.i, sys/inc/rmunderover.i


brad-advantzware (changes requested):
activates the log value on NK1 but does not meet specifications.  See ticket

brad-advantzware (approved):

PR #2759, opened 2018-06-01
Ticket - 29836 N-K-1 BolFmt = Soule & Soule Med Shipto Notes from Customer Shipto File are Disappearing from BOL

sewa67:
Files - oe/rep/bolsoule.p


brad-advantzware (changes requested):
Sewa Singh The "Soule" format grays out the Print Ship Notes option.
The "SouleMed" form has that option active, yet it prints ship Notes regardless.
Your change makes the ship notes print regardless. The better change) is to have the check boxes active for both formats and have the ship notes print if the check box is checked and not print if the check box is not checked.

brad-advantzware (approved):

PR #2761, opened 2018-06-01
29866 Selecting 'OK' to Quote or BOL print selection parameters when the format is not available, knocks user out

sewa67:

Files - est/r-quoprt.w


brad-advantzware (changes requested):
I increased the scope for this and changed so that this is encapsulated

brad-advantzware (approved):
This is an odd and very specific way to prevent an error on Quote form only.  Merging for testing.

PR #2763, opened 2018-06-01
#25634 Menu Option for - Touch screen Clock In/Clock Out

tyndmar:


brad-advantzware (approved):
shownote.i change just for compile purposes?  doesn't seem related to ticket

PR #2775, opened 2018-06-05
Ticket - 18883 Add Due Date and Print Vendor Header option to VR1

sewa67:
Files - ap/ap-agedN.i,aprep/r-vagingN.w


brad-advantzware (approved):
Will need to be validated with testing.  Can't confirm this will work through code review

brad-advantzware (changes requested):
Strike that, from image, this will fail because format is cut off on header lines

brad-advantzware (approved):

PR #2777, opened 2018-06-06
Ticket - 30052 O-U-10 issue in version 16.6.12 - Rel Lot# not tied to same Release as PO#

sewa67:
FIles - oerep/r-casetg.w


brad-advantzware (approved):
Odd that an output parameter was already there but not used

PR #2790, opened 2018-06-11
Ticket - 30497 + Sign COPY Button Bug on Customer Part# on two Forms.Update Button has Same Error Message.

sewa67:
Files - ce/b-estitm.w,ce/b-estitm1.w,cec/b-estitm.w


brad-advantzware (changes requested):
Sewa - 
This appears to me that it will bring back the problem with a 2 Pc box requiring a different part # between form 1 and form 0.  The warning must only be presented when the part # appears more than once *within* the same .form-no.  It *must* not appear when the duplicate part# is not on the same form.

brad-advantzware (approved):
Sewa - I will approve this based on your request but I don't understand how stripping out these checks is fixing the problem.  I will approve to testers.

PR #2792, opened 2018-06-11
Ticket - 30501 WHATIF Button must FOCUS on Field

sewa67:
Files - browsers/probe.w


brad-advantzware (approved):
Unsure why/how a full copy of the browser object was added to this file but this change looks to delete the whole object.

PR #2800, opened 2018-06-12
Ticket - 28407 Raw materials issue with an invalid tag number was accepted and processed

sewa67:
Files - rm/r-rmte&p.w,sys/inc/create_nk1.p


brad-advantzware (changes requested):
Need to allow for blank tag.  should just add: 
AND rm-rctd.tag NE ""

to the joined FIRST query of rm-rctd

brad-advantzware (approved):

PR #2808, opened 2018-06-14
28764 - Import Enhancements - Support option to Ignore Blanks

brad-advantzware:


tyndmar (approved):
Not sure about the technical details, but this reads as compilable and logical

PR #2813, opened 2018-06-14
Ticket - 28698 Margin % Column

sewa67:
Files - browsers/probe.w


brad-advantzware (approved):

brad-advantzware (changes requested):
Sewa - I think I would prefer to remove the Margin% column from the browser, not just make it hidden.  Why did you choose to make it Visible=FALSE instead?

brad-advantzware (approved):

PR #2830, opened 2018-06-18
30915 Order Item Price Matrix Error

wade-advantzware:


brad-advantzware (commented):
Wade - Is it possible to assign the values to the oe-ordl first prior to calling this?  I thought the call was made after the local assign.  Does that not fill in the pending values on the line?

brad-advantzware (approved):

PR #2833, opened 2018-06-19
#30102 GF2 Export

jgill-advantzware:


brad-advantzware (changes requested):
getValue-itemfg and the BUFFER variable name are misleading and should be changed.  This was not introduced but this should be corrected to "getValue-account" and the Buffer variable to "ipbf-account".  This was likely copy/pasted from the IF1 export and never changed.

PR #2835, opened 2018-06-19
24799 Invoice Edit Error

wade-advantzware:


brad-advantzware (commented):
Wade - Why are some nk1 .i programs moved into a DO TRANSACTION and others not?  Comments in code would help explain the change that you made.  I like that comments were added to ENDs but in b-ordm.w and oe-bolp7.p, I'm not clear on the reasons for these changes

brad-advantzware (approved):

PR #2837, opened 2018-06-19
30953 UDF GRN field issue

ron-advantzware:


brad-advantzware (approved):
Ron - What change elsewhere introduced the use of the "|" in the data?  Is there a source ticket for that?

PR #2839, opened 2018-06-20
30975 Add Version to Module Headers

ron-advantzware:


brad-advantzware (commented):
@ron-advantzware You requested a closing of a commit for this.  But is this one still good?  Question here is why is the sysconst.i not added to the winMethods.i directly?  Isn't there a chance that the version constant is not defined?

brad-advantzware (approved):

PR #2841, opened 2018-06-20
Ticket - 31001 Error: No BF -ItemFg Record is Available when adding Order for Combination Estimate 13042.

sewa67:
Files - oe/priceprocs.p


brad-advantzware (approved):
This will suppress the error message but the issue likely lies upstream from this call, in the assignment of the oe-ordl record.  Approving this and assigning to Wade who was looking into a similar problem

PR #2853, opened 2018-06-21
Ticket - #30048 Valley Loftware load tags not printing properly

sewa67:
Files - oerep/r-loadtg.w


brad-advantzware (changes requested):
removeChars surround i-name and i-no but not the new description field.  Should be consistent with use of removeChars

brad-advantzware (approved):

PR #2857, opened 2018-06-21
28759 Improvements to Main Menu 2 to enhance automated upgrades

ron-advantzware:


brad-advantzware (changes requested):
No more .i for NK1s.  Add a "Case" to Create_NK1 and use nk1look instead.

PR #2866, opened 2018-06-23
Ticket - 31211 PO issue when trying to enter a PO

sewa67:
Files - po/d-poordl.w


brad-advantzware (commented):
Ideally, I think this should be managed by tab order and enable/disable.  @ron-advantzware Do you see a better way to do this than doing "apply entry" all over the place rather than just naturally progressing through tab order?  Can tab-order be changed dynamically?  You can approve this code review but if you think it can be done better, please create ticket to refactor for future consideration.

PR #2871, opened 2018-06-25
Ticket - 30954 Add First Resource (Machine) to Purchase Order Screen

sewa67:
Files - po/d-poordl.w


brad-advantzware (changes requested):
Code to determine first machine should be a sub procedure

brad-advantzware (approved):
OK, but I would prefer the sub procedure be parameterized. rather than use screen-value (separate UI from business logic).  Approving to get this tested.  This should be handled in PO rewrite

PR #2881, opened 2018-06-27
Ticket - 19315 POPrint=Peachtree - Add First Machine.

sewa67:
Files - po/po-pchtree.p


brad-advantzware (changes requested):
If v-vend-i-no is blank and first machine is not blank, the blank vend-i-no will print.  Am I correct that this OR cMachCode NE "" should no longer be a condition to print he v-vend-i-no

brad-advantzware (approved):

PR #2891, opened 2018-06-27
Release/16.8.0   working copy

jgill-advantzware:


brad-advantzware (changes requested):
The fi_width-min, fi_length-min, fi_width-cst, fi_length-cst should not longer be necessary.  These should be converted to use the table fields directly.  This would also allow you to remove "reftable-values" procedure completely.

brad-advantzware (approved):

brad-advantzware (approved):

brad-advantzware (approved):

PR #2900, opened 2018-06-28
Feature/27443 auto log off

wade-advantzware:


brad-advantzware (changes requested):
As discussed this morning.  Logic for what is controlled by Admin vs. Super-admin should use the PgmMstrSecur.p

brad-advantzware (changes requested):
Thanks Wade - I think this pushes the logic down where it should be.  One thing, we don't want to match on user ID "ASI".  The lIsAnAdmin should be enough.  (we might change the "super-user" account and don't want to change code.)

brad-advantzware (approved):

brad-advantzware (commented):
Note @mark-advantzware potential merge conflict.  Please make sure the small change for main menu ("CanUpgrade") makes it in to the merged code

PR #2901, opened 2018-06-28
31032 ASI cannot log in

tyndmar:


brad-advantzware (commented):
I guess it is good to simplify but where are you moving logic to assign aliases and manage iLockoutTries?  Was that duplicated somewhere else?

brad-advantzware (approved):

PR #2905, opened 2018-06-28
31710 EDI 810 Invoice Report

wade-advantzware:


brad-advantzware (approved):
Approving but not sure how this relates to the ticket.  Is this change just to remove a loop that will never result in a useful entry in the exception log?

PR #2907, opened 2018-06-28
31572 Create BOL button update release message

wade-advantzware:


brad-advantzware (approved):
Approving but is there a need to empty the temp-table?  Any chance this could be run persistent or hang-around after open/close of the window?  I guess worse case is that they don't get warned and keep scanning a release to the DB (which we aren't certain is a bad thing, correct)?

PR #2909, opened 2018-06-28
31013 Indiana Carton SB Job Scan feedback

ron-advantzware:


brad-advantzware (approved):
/* scan has leading +, remove it */
-                cResource = REPLACE(cResource,"+","").
+                cResource = REPLACE(cResource,"-","").

Why is there a symbol in their scan?  What is the supported scan format for the system?  Worried that this is very proprietary

PR #2916, opened 2018-06-29
31747 Error upon launching Order Entry: Field managed from oe-ordl re…

tyndmar:
…cord

re-added field to FIELDS-IN-QUERY preprocessor


brad-advantzware (approved):
App builder required adding to Fields phrase and adding to browse (but hiding)

PR #2927, opened 2018-06-30
31857 Update Resources/StdMenu files from working source

tyndmar:


brad-advantzware (approved):
@mark-advantzware I see a couple "clayA" and "asi" references in the prmgs.d  Are these ignored in the loader?
Approving pending your review of that

PR #2930, opened 2018-06-30
31866 Modify asiUpdate files to support new version number

tyndmar:


brad-advantzware (commented):
Use o Value(cAudName) replaced with hard-code of audit.  is this desired?
What are typical values of ipiLevel passed to this?  Is "asi" GT 10 and non-admin LT 10?

PR #2932, opened 2018-07-02
Ticket - 25353 <NEW> NK1-BOLFmt=BolFMT15- copy of Xprint2 w' business formlogo at top and units at the bottom

sewa67:
Ticket - oe/rep/bolfibrex.i,oe/rep/bolfibrex.p


brad-advantzware (changes requested):
Sewa - This is still just changing the BOLFMTX format for FIBRE.  Before you make anymore changes, we need to clarify spec

brad-advantzware (approved):
THis is a mess but I think the BOLFMTX with "BOLFMTX15" as the char value, should print the Freight BOL as desired.  Approving to test but will need to fail other requirements (including BOLFMT15 for BOLFMT which appears to call a non-existent program).

PR #2948, opened 2018-07-06
Ticket - 22216 Network Utilities Plan.

sewa67:
Files - util/b-utilsN.w, util/util-exp.w, util/utilsN.w, util/v-utilsN.w


brad-advantzware (approved):
Note - Approved to merge for now (Per Jay, he would like to test this if possible).  These new .w files do not open within App Builder thus this would fail review, so after compile, just let Jay know that it can be tested, but I will flag ticket as Failed Testing for Sewa to resolve the App Builder issues.

PR #2980, opened 2018-07-11
#25482 Reftable - AR-CASH

jgill-advantzware:


brad-advantzware (changes requested):
with ar/r-dbcr&p.w, instead of the "if ar-cash.printed ne YES then NEXT and if ar-cash.stat eq 'H' then next, convert these to criteria on the  For Each loop.

Missing ar-cash.printed = YES in ar/rep/crdbmemo.p

PR #2987, opened 2018-07-12
Ticket - 22216 Network Utilities Plan

sewa67:
Files - system/PgmMstrSecur.p,util/b-utilsN.w,util/utilsN.w


brad-advantzware (changes requested):
Sewa - I don't see where it checks for the security level when running.  The code to execute a utility should be the one procedure, regardless if run from browser or viewer.  The check for existence is correct but a message should be shown if the user doesn't have a security level high enough to run the utility.

brad-advantzware (changes requested):
Must not have test of security level LE 999 outside of PgmMstrSecur.p

brad-advantzware (approved):

PR #2990, opened 2018-07-12

ron-advantzware:
…rt require license to be enabled


brad-advantzware (changes requested):
Ron - I may be wrong but I don't see the call to PgmMstrSecur from within CheckModule to bypass the module check for super-admin

PR #2992, opened 2018-07-12
#25483 Reftable - ar-cashl.return

jgill-advantzware:


brad-advantzware (changes requested):
Legacy/salrep/createTempInvCash.p - Proc getoeret changed parameters but callers of getoeret still passing 3 parameters.  This will compile but give a run time error.
Additionally, the procedure getoeret.p changed parameters.  This is a FUNDAMENTAL problem where there callers to this .p were not included in this code change.  This will result in mismatched parameter run times and is indicative that you did not test this.

PR #2993, opened 2018-07-12
Feature/29197 implement incoming sheet orders for valley

wade-advantzware:
Changed file mask from * to *.dat


brad-advantzware (approved):
Currently already deployed but monitor not running.  Approving with conditions:
1) Create ticket to refactor the InboundFTP to one program that encapsulates all WinSCP functions (parameterized)
2) WinSCP.ini should not likely be tracked or we need to understand what all .ini changes are for and what we should distribute as customer template

PR #2998, opened 2018-07-16
#25487 Reftable - cp-lab-p

jgill-advantzware:


brad-advantzware (changes requested):
This all seems completely wrong.  Why are there log fields instead of character fields?

PR #3001, opened 2018-07-17
Ticket - 24094 N-Q-2 Update Customer Code / Merge Fails to Update Ship To on O-U-1

sewa67:
Please review this change and merge into development 
Files - util/updcust#.w


brad-advantzware (changes requested):
Test for b-cust.active eq "I" is not correct.  Conversion should not change inactive to active.  the logic should just look for blank (.active eq "") and set this to "A"

brad-advantzware (approved):

PR #3004, opened 2018-07-17
#25668 Reftable - PREPCADFILE

jgill-advantzware:


brad-advantzware (changes requested):
viewers/prep.w - why are there separate fill-ins kept for this.  They should be converted to native ADM DB fields.  Do that for last-job-no and last-job-no2 as well.

PR #3006, opened 2018-07-17
31864 Update user table screen - removing values that we do not want …

ron-advantzware:
…to use


brad-advantzware (changes requested):
Use the function instead of hard-coding the 900 and 999 in code.

PR #3009, opened 2018-07-18
Ticket - 29453 P-U-3 and P-U-1 Icon to Print Purchase Order, Output to Email Subject Line should include the PO

sewa67:
Files - poprep/r-poprt.w


brad-advantzware (commented):
@wade-advantzware  Can you look at this one.  I don't want copy/paste in code and that appears to be what was done here by Sewa.  I feel like the original query can just add a Break by and then a conditional FIRST-OF(po-ord.po-no) rather than a whole separate code block.  Can you give Sewa some guidance on this?

brad-advantzware (approved):

PR #3014, opened 2018-07-19
29286 - reftable1 fixes

tyndmar:


brad-advantzware (changes requested):
Think we should remove the code that references ASIHelpServices (or whichever one is not correct)?

PR #3015, opened 2018-07-19
31950 Link on Main menu needs to be changed for all users upon upgrade

tyndmar:
updated asiUpdate to force NK1 value


brad-advantzware (approved):
Why are all the reftable1 (oe-rel.lot-no) changes also in this commit/pull request?

PR #3016, opened 2018-07-19
31292 Estimated list to Excel

tyndmar:


brad-advantzware (changes requested):
Mark - Shouldn't this replace the existing text based criteria within the query?

brad-advantzware (approved):

PR #3030, opened 2018-07-21
Ticket - 28633 Transfer BOL - Will Not Post when destination is same as current loc.

sewa67:
Files - oe/r-bolpstN.w


brad-advantzware (changes requested):
Do not want pop-up message in posting.  Giving back to prod-dev to review request

PR #3034, opened 2018-07-23
#25447 Reftable - ORDERPO

jgill-advantzware:


brad-advantzware (changes requested):
Jitender - Set up some time to review this with me tomorrow

brad-advantzware (changes requested):
See inline comments.  Fix converter.  Please summarize how the rowid passed to FindExistingPO is changed to a job-mat rowid.

PR #3038, opened 2018-07-24
Ticket - 32138 NK1-POPrint=LancoYork - format not correct.

sewa67:
Files - po/po-lanyork.p


brad-advantzware (changes requested):
The issue with the scores is that you are passing decimals in the form of 16.18 (4/16ths) into the fraction converter. You need to either convert these to true decimals (16.18 = 16.5) or capture the .18 and divide it by the customers setting for CESCRN (in AWS this is 32nds). There are functions to handle these conversions (sys/inc/k16)

PR #3043, opened 2018-07-24
33020 Tri-lakes - extreme slowness of application

tyndmar:


brad-advantzware (commented):
@mark-advantzware This ticket is closed but has this open PR.  Should this change have been merged?

PR #3044, opened 2018-07-24
30971 TS machtran creation rewrite

ron-advantzware:


brad-advantzware (approved):
Approving.  More sub-procedures would be better and procedures should be named with the function that they perform.  Job_Data_Collection_2 is not a good proc name but at least it breaks up the monolithic procedure a little.

PR #3046, opened 2018-07-24
25438 Reftable - est.calcpcts

jgill-advantzware:


brad-advantzware (changes requested):
@jgill-advantzware Are you sure that xest will have an assignable LOCK status when values get assigned to it?

PR #3049, opened 2018-07-25
Ticket - 24094 N-Q-2 Update Customer Code / Merge Fails to Update Ship To on O-U-1

sewa67:
Files - util/updcust#.w


brad-advantzware (approved):
Not a big fan of disabling triggers but this is likely the best short term fix (and other triggers are already disabled)

PR #3053, opened 2018-07-25

ron-advantzware:
…rt require license to be enabled


brad-advantzware (changes requested):
Won't this fail?  The line in PgmMstrSecur.p is:
(ipProgName EQ "util/module.w" AND ipFunction EQ "") OR /*run util module*/
The ipFunction needs to be "", no?

PR #3055, opened 2018-07-25
#25439 Reftable - est/getqty.w2

jgill-advantzware:


brad-advantzware (approved):
Approved.  This may be unnecessary to store this field, but reftable replacement looks OK.

PR #3057, opened 2018-07-26

ron-advantzware:
…rt require license to be enabled part 2


brad-advantzware (changes requested):
Now you are breaking the original call from util/module.w...
You had it correct in the first place.  Just add a "CheckModule." entry to PgmMstrSecur.p.  Assign to me if necessary

PR #3058, opened 2018-07-26

ron-advantzware:
…rt require license to be enabled part 3


brad-advantzware (changes requested):
You removed the ability for super-admin to edit modules using util/module.w.  I'm taking this.

PR #3059, opened 2018-07-26
32460 SuperAdmin CheckModule

ron-advantzware:


brad-advantzware (changes requested):
Add line, don't replace util/module.w

PR #3062, opened 2018-07-26
Ticket - 32613 N-K-1 SSPostFGXfer Logical Value = YES will Post Transfers Immediatly after Saving the Transaction Line

sewa67:
Files - addon/fg/b-trans.w,panels/p-updba2.w,sys/inc/create_nk1.p


brad-advantzware (changes requested):
We can't have the fg-posting procedure copy and pasted in.  I know it is like that in other SS applications but it needs to be the same procedure.

brad-advantzware (approved):

PR #3071, opened 2018-07-27
28737 Cannot Import Correct Machine on Estimate

sewa67:
Files - cec/mach-seq.p,est/MachLimitsDims.p,est/d-machex2.w


brad-advantzware (changes requested):
I don't think we want to convert the decimal display in these messages.  We may be closing this pull request without merging.

PR #3075, opened 2018-07-27
#25462 Reftable - ShiftDays

jgill-advantzware:


brad-advantzware (changes requested):
Conversion needs to set .useSpecificDays based on the value of .loc in reftable ("1" = YES, otherwise, NO).
The If avail Shifts block in the viewer does not seem necessary.  You need to make sure that shifts is writable at that point in the code (if adm, it should be but test)

PR #3076, opened 2018-07-27
#25466 Reftable - msf-limit

jgill-advantzware:


brad-advantzware (changes requested):
convert the fi_msf-limit into native truck.msfLimit field and remove all extra non-adm code

PR #3077, opened 2018-07-27
33146 Error in PO Tab of IF1

ron-advantzware:


brad-advantzware (changes requested):
Message left in itemfgpo.w - Also, I'm very uneasy about modifying var.i and about using a global variable inside a pop-up browser.  It may work but I guess I need to be convinved

PR #3085, opened 2018-07-29
33023 - H-R-1 Not showing all data

tyndmar:
Commit 1 - ONLY cosmetics, NO logic changes


brad-advantzware (changes requested):
DEF VAR list-name AS CAHR NO-UNDO.
Also, why are full expanded keywords reduced?  That is not per policy and if you use Progress Dev Studio, is part of the standard beautification that we want as per coding standards.

PR #3089, opened 2018-07-30
Ticket - 33090 Need to gray out the approved button.

sewa67:
Files - oe/vp-ivhld.w


brad-advantzware (changes requested):
Why is there extra code in the local-initialize that appears to be from Yoosun?

brad-advantzware (approved):

PR #3093, opened 2018-07-30
27235 Ability to view NK1 Values within a Module

ron-advantzware:


brad-advantzware (changes requested):
Ron - I still see "AdminOnly" reference in here.  I thought we confirmed that we do not need to add this field to the sys-ctrl table?

brad-advantzware (approved):

PR #3095, opened 2018-07-31
#25464 Reftable - SPECSAMP

jgill-advantzware:


brad-advantzware (approved):

brad-advantzware (changes requested):
Need to release the exclusive-lock on the est-table within this report.  I worry about transaction scope with this.  Adding to standup for recommendations for how to manage transaction scope best.

brad-advantzware (changes requested):
Must do a FIND CURRENT est NO-LOCK after assigning to it.
Also validate that RELEASE sys-ctrl is ok and that the sys-ctrl buffer table doesn't get used.

PR #3096, opened 2018-07-31
33348 Fix to handle shift spanning midnight

ron-advantzware:


brad-advantzware (approved):
Approving for expediency.  Assuming the loop continued unwanted without this new condition

PR #3106, opened 2018-08-01
24669 Scanning that creates backorder

wade-advantzware:


brad-advantzware (commented):
Wade - In your tests, were they on completely different orders?  It seems like the ticket talks about different order lines within the same order.  Shouldn't the system try to match on .line and .i-no as well?

brad-advantzware (approved):

PR #3114, opened 2018-08-03
Ticket - 33479 Error message when editing the set header on a partition.

sewa67:
Files - cec/d-updsetpart.w


brad-advantzware (changes requested):
Check for eb.stock-no ne "" before you do the find on the itemfg.  You want the FIND CURRENT itemfg NO-LOCK to run regardless of eb,stock-no ne ""

brad-advantzware (approved):

PR #3145, opened 2018-08-16
Ticket - 33830 IF1 Import Question for FG Items requiring a customer code.

sewa67:
Files - viewers/itemfg.w

### DIFF
--- a/Legacy/create.trg/itemfg.p
+++ b/Legacy/create.trg/itemfg.p
@@ -36,6 +36,12 @@ FIND FIRST fgcat
     WHERE fgcat.company EQ g_company
     USE-INDEX fgcat NO-LOCK NO-ERROR.
 
+FIND FIRST cust no-lock
+             WHERE cust.company EQ g_company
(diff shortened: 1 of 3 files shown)


brad-advantzware (changes requested):
This will likely be aborted.  See ticket for revised requirements

brad-advantzware (changes requested):
create.trg not edited as described in ticket.  I don't want default logic in the UI

brad-advantzware (changes requested):
Sorry for the misdirection on this but the code should go into create.trg/itemfg.w (not write.trg).
This needs to apply the customer X as default and *then* assign the .cust-no from the FGMaster record if it is not blank.
Additionally, the viewer should not prompt for anything related to the .i-code eq "S". Remove that entirely and only keep the prompt protecting against the Customer code being blank.
You can remove the code in the write.trg that you added as well. Look for my notes below

brad-advantzware (approved):

brad-advantzware (approved):

PR #3146, opened 2018-08-17
Ticket - 33463 Request - Adding UOM to IC1

sewa67:
Please review this change and merge into development  branch 
Files - fgrep/r-cycntN.w


brad-advantzware (approved):
Note:  Cost UOM was not implemented but approving since the cost is always represented in Cost/M.

PR #3149, opened 2018-08-20
32152 New NK1 v4

ron-advantzware:


brad-advantzware (approved):
Approving to get into hands of testers.  Will want to look closer at CRUD operations, but OK if not "framework-level" yet.

PR #3152, opened 2018-08-20
Ticket - 28567 Change to O-R-5

sewa67:
Files -oerep/r-bookedN.i,oerep/r-bookedN.w


brad-advantzware (changes requested):
I would prefer these be in a common program that returns: "GetRoutingForJob"  "GetInksForJob".  Also, I don't see the AOA version of OR5 changed.

brad-advantzware (approved):

PR #3164, opened 2018-08-21
#25492 Reftable - POORDLDEPTH

jgill-advantzware:


brad-advantzware (changes requested):
I think all code lines where you do IF AVAIL po-ordl then po-ordl.s-dep before the other conditions to use the item.s-dep should be changed to 
If po-ordl.s-dep GT 0 then po-ordl.s-dep
Otherwise, the other conditions will never be met since po-ordl is always available at that point (already assigned to v-wid and v-len)

brad-advantzware (approved):

PR #3167, opened 2018-08-22
Ticket - 34285 SHT CALC Button Bug shows No Lock. This.

sewa67:
Files - jc/shtcalc.p


brad-advantzware (approved):
Not an ideal approach of managing an un-locked buffer.  But I don't think this causes any new locking issues and seems to solve the reported problem

PR #3169, opened 2018-08-22
Ticket - 34263 N-K-1 OEAutoPO via J-U-1 Not Converting Purchase UOM

sewa67:
Files - po/dopo.p


brad-advantzware (changes requested):
Sewa Singh I do not understand your change. Can you clarify with an example?
Why does the material type need to equal "B"?
The conversion needs to be from "EA" (sheets) into the "Qty Purchased UOM" (item.pur-uom). So in Joe's example, the PO should come up as
Quantity: 28,742.0 LF after converting Sheets into LF.

brad-advantzware (approved):

PR #3181, opened 2018-08-27
Ticket - 34158 Estimate Set Header Count Changed after Update

sewa67:
Files - cec/b-estitm.w


brad-advantzware (changes requested):
Sewa - the call should require the "adm-new-record" to be true to run the copy-2-form-zero (in addition to the criteria already there)

brad-advantzware (approved):

PR #3188, opened 2018-08-28
Ticket - 34263 N-K-1 OEAutoPO via J-U-1 Not Converting Purchase UOM

sewa67:
Files - poinq/b-po-inq.w


brad-advantzware (changes requested):
This will likely not work and never display information for purchased finished goods.  The correct fix here is to properly set the po-ordl.t-rec-qty and po-ordl.cons-uom upstream from how it is displayed in the browse.  Recommending to close this pull request and defer this defect.

PR #3195, opened 2018-08-29
29642 I-U-2 Transfers from Warehouse to Another Warehouse is changing…

tyndmar:
… Cost

cost-uom value not being set correctly during txfr order


brad-advantzware (changes requested):
Mark - I don't think this will work (the ext-cost calculation is still dependent on the pur-uom) but I'm not even sure this is where the problem is.

PR #3196, opened 2018-08-29
#25463 Reftable - d-shp2nt

jgill-advantzware:


brad-advantzware (changes requested):
Still need conversion program.  Do not need to convert reftable to user-print.  just remove reftable (to reftable1)

brad-advantzware (approved):

PR #3199, opened 2018-08-29
30703 N-K-1 FGRecept Logical Value = Yes

wade-advantzware:


brad-advantzware (changes requested):
No need to manage the actual password nk1 request.
Need to move password prompt out of get-job-no and into own func/proc.
Simplify logic of FGRECPT function and enclose it in one function/proc.
valid-job-no does most or the same stuff.  no duplication.

brad-advantzware (approved):
Unsure if tied to the correct UI triggers.  Will need to be tested so approving.

PR #3205, opened 2018-08-30
#25481 Reftable - ar-cashl.inv-line

jgill-advantzware:


brad-advantzware (changes requested):
There is a delete reftable that is not converted
(If NOT tt-inv.selekt then delete reftable).  I do not feel comfortable with this change and recommend deferral to 16.8.4

PR #3212, opened 2018-08-31
Ticket - 34810 Cannot post AP checks in version 16.8.0.

sewa67:
Files - ap/r-chkreg.w


brad-advantzware (changes requested):
See ticket for feedback. I don't think this change should be made.

PR #3226, opened 2018-09-04
#25488 Reftable - ChargeCode

jgill-advantzware:


brad-advantzware (changes requested):
Jitender - the fgcat.w viewer should just convert the v-charge, v-gl-rm, and v-gl-fg into native database fields that behave exactly the same as fgcat.glacc

PR #3229, opened 2018-09-05
Ticket - 30911 Error when Items are Issued to a job..

sewa67:
Files - rm/rm-autoi.i, rm/rmpost.p


brad-advantzware (changes requested):
What is the purpose of this code?  Why is it different than the code in

brad-advantzware (commented):
Closing this pull request.  Just using an Avail should be fine and consistent with the main posting program.

PR #3235, opened 2018-09-06
Feature/32459 releases reference non existent shiptos

wade-advantzware:


brad-advantzware (commented):
What is "And shipto.ship-no ne 1"  ?  Need comment on why that is there...

brad-advantzware (commented):
Wade - I commented in ticket.  After more thought, I think this still needs to be changed.  I don't think the NE 1 approach is correct but I also don't think the Ship ID will always be valid when accompanied with the BOL customer (transfer releases within standard orders)

brad-advantzware (approved):

PR #3236, opened 2018-09-06
33385 Numerous FG Pending Receipts at Peachtree

wade-advantzware:


brad-advantzware (approved):
@wade-advantzware Approving. Just to summarize your change:  You removed all copied code of the posting program out of these sharpshooter programs and replaced then with a call to fgPostBatch. The additional changes were just to include all tts and variables needed in the fgPostBatch calls.

PR #3237, opened 2018-09-07
Ticket - 34933 J-L-8 report is not on the standard menu and columns are cut off the report

sewa67:
Files - jcrep/r-jcstdN.w


brad-advantzware (changes requested):
Sewa - The change to Resources/StdMenu is missing here.  Has that already been done?

brad-advantzware (changes requested):
These changes to the menu files should be a single additional line.  These should all be rejected.  I will assign to Ron because we are changing the way that we define menus anyway.

PR #3244, opened 2018-09-10
Ticket - 29997 Request - Reason Codes for Adjustments

sewa67:
Files - fg/b-fgadj.w,fg/fgbjobu.w,fgrep/r-fgpstrN.i,fgrep/r-fgpstrN.w,oeinq/b-ordfgi.w,rm/b-rmadj.w,rm/d-rm-bin.w, rminq/b-rmiinq.w,rminq/d-rmiinq.w,rmrep/r-trans2.i,rmrep/r-trans.i,rmrep/r-transN.w,sys/inc/create_nk1.p


brad-advantzware (changes requested):
Build a RejectCodeProcs program so that you write the code to validate a reason code (can encapsulate the NK1 lookup too), return the reason description, build the combobox string, etc.  Same code is copied in multiple places.  You can make it accept a .type so that other consumes of the rejct-cd table can be converted later.
Code to assign to the history records should not be in window.  The value should be added to the "A" transaction that gets created from the inputs provided in this window.  For instance rm/cre-tran needs an additional argument for reason code that needs to be captured from the user in the window.  Same for the FG/fgbjobu.w 
Where are the changes to the posting program to propagate the fg-rctd.reject-code to fg-rdtlh.reject-code?  Same question for rms?

brad-advantzware (changes requested):
Sewa - For the RM adjustment made through MF1, finding an "A" transaction in history within d-rm-bin does not seem like it will work.  If you look at the caller (rm-ibin.w) the "A" transaction is created *after* a change is made in rm/d-rm-bin.w.
I think you will need an output parameter added to d-rm-bin that includes the adjustment code.  Then you need to add an input parameter to rm/cre-tran.p to pass that variable in when creating the history transaction.

SImilary for For FG:  there is a fg/cre-pchr.p that creates the "A" record (this is when the OK button is pressed).  

All the code that does the query for fg-rdtlh inside these programs should be removed in favor of using the cre-  procedures to create the history record and assign the reason code.

brad-advantzware (approved):

PR #3266, opened 2018-09-13
Ticket - 35448 Prep/Route tab - MR Rate column cuts off 3 digit ratesin Operations browser.

sewa67:
Files - ce/b-estop.w, est/b-estop.w


brad-advantzware (approved):
Sewa - I can approve but why are the columns between ce/ and est/ different widths?

PR #3275, opened 2018-09-14
Ticket - 35516 O-F-2 Toggle Box Displays as Unchecked but is actuallychecked.

sewa67:
Files - custom/oe-ctrl.w


brad-advantzware (changes requested):
Sewa - Isee a duplicate line above your Do with Frame Block.  What is the purpose of this additional assigned to tgCreateSSBOL:screen-value?  
There shouldn't be any additional Enable/Disable that is different from the rest of the controls.  There is also another DISABLE tgCreateSSBOL.
Can you clean this up too?

brad-advantzware (approved):

PR #3283, opened 2018-09-15
Ticket - 22216 Network Utilities Plan

sewa67:
Files - system/PgmMstrSecur.p,util/custproc.w,util/b-utilsN.w,util/utilsN.w


brad-advantzware (changes requested):
What is query-first code? This looks to be copied code querying the "custbrowse" NK1 for # of records.  This should not be in this file.  What else was copied over?

brad-advantzware (approved):

PR #3290, opened 2018-09-17
35607 Remove references to translations.txt in Touch Screen and use M…

ron-advantzware:
…enu database structure


brad-advantzware (commented):
Ron Stark Some questions:
Why is there a translation field on prgrms?
I think it would be better to name the temptable ttTranslations so that it is more clear that this is a temp-table. tougher to read with the translation (no s) table.
Why is this in a .i ? When this converts to golden code, this code will be repeated in Java for every instance of this .i. IF we have a program that we call, only that program/proc gets converted into a modular package. Clear input-output. The answer is probably just because it was already a .i but I think we should reserve .i for definitions of temp-tables only. Also, if we change this, we have to compile all consumers.

brad-advantzware (approved):

PR #3292, opened 2018-09-17
35606 Main Menu Enhancements

ron-advantzware:


brad-advantzware (approved):
Approving - General suggestions:
transalations.i -> translations.p that encompasses the other addon translations.i
Use "sfGetUserSecurityLevel"  Get is a common OO way of pulling a "property"
Comment per line rather than comment block (or just delete the code)
Try to use all sub proc calls in Main Block if possible.  Easier to read.

PR #3303, opened 2018-09-19
Ticket - 26799 Request - EQ Quotes Import/Export, Add expiration date Add status

sewa67:
Files - est/b-qthd.w,est/v-qthd.w,est/w-quote.w,oe/d-oeitem.w,oe/d-ordqty.w,
           oe/d-quotedprices.w,util/validate.p,util/ttImport.i,utilImportQuote.p,util/ValidateCustPart.i,
           util/validateest.i,util/impQuote.p


brad-advantzware (changes requested):
Sewa - Generally looks like a good job.  
Few changes:
When testing the .expireDate you also have to allow ? in that field.  No expiration means never expires.
The Import comment header is copied from ImportCustomer.  Update that to be accurate.
Why is the "iplUpdateDuplicates" disabled?  That should be managed just like other Update-able import programs.
There should be a pAssignValueDt, if not create one and use it to handle assignments for date fields.

brad-advantzware (approved):

PR #3313, opened 2018-09-20
35590 Remove itemfg spare-char-5 Stack Trace

ron-advantzware:


brad-advantzware (changes requested):
Remove the variable cPgmStack and the code that assigns it too.

PR #3325, opened 2018-09-23
35705 Restore F3 to New NK1

ron-advantzware:


brad-advantzware (approved):
I don't think the DB Connection code is required anymore but approving since that is not a critical change at this point

PR #3330, opened 2018-09-24
36086 Add Import and Export functionality to NK1

ron-advantzware:


brad-advantzware (approved):
Approving although we are not going in this same direction.  Ticket will be reassigned to Sewa to build Export in Import-ready format.

PR #3336, opened 2018-09-25
22929 Order not pulling correct information

wade-advantzware:


brad-advantzware (changes requested):
Same block of code within the same program -> Function
GetPreviousOrder

tyndmar (approved):

PR #3343, opened 2018-09-25
additional task for deleting "child" tags

tyndmar:


brad-advantzware (commented):
@mark-advantzware I don't think we should bother matching on linktable.  rec_key should be unique so just delete all children based on rec_key.  Also, why do we need CAN-FIND here?

PR #3347, opened 2018-09-25
35609 Place generation of rec_key value in session.p

ron-advantzware:


tyndmar (approved):

brad-advantzware (commented):
@mark-advantzware @ron-advantzware Shocking to see how many places outside of the create.trg that assign rec_key - creating ticket 36245 for future handling but good to see this as a function

PR #3366, opened 2018-09-27
Ticket - 35042 Required Fields for AF1 Import.

sewa67:
Files - util/ImportCust.p


brad-advantzware (approved):
Not sure I agree but this is coded per direction from Clay.  Approving for testing.

PR #3369, opened 2018-09-27
32481 - additional mods, simple phone/email

tyndmar:


brad-advantzware (changes requested):
Remove last 3 fields from viewer (the two "Chg" fields and Active

brad-advantzware (approved):

PR #3370, opened 2018-09-27
36052 Utilize users Mnemonic settings within Main Menu

ron-advantzware:


brad-advantzware (approved):
This was a large diff but from what I can see, you added code to pull changed values from the user file into the Main Menu user settings.
Also assuming the delete of mainmenu.i was intentional?
(Please confirm) @ron-advantzware

PR #3372, opened 2018-09-27
26325 Estimate Quantity Screen - Fixes

brad-advantzware:


tyndmar (approved):
A little tough to sort the logic from the cosmetics, but I don't see any blockers.  Approving.

PR #3377, opened 2018-09-28
32481 - Added F1 lookup to default bin field

tyndmar:


brad-advantzware (approved):
Seems OK but not sure we want a "Default Bin" when there is no functionality for it.

PR #3378, opened 2018-09-28
35628 - add location record creation for existing loc records

tyndmar:


brad-advantzware (commented):
Mark Tyndall The code you just added for this.  How can we be sure this is run once?  Should the utlity be safer and just go through each loc and first see if location exists, then add it, rather than deleting all location records first?

brad-advantzware (approved):

PR #3393, opened 2018-09-28
35645 - Taxable Defaults

brad-advantzware:


tyndmar (approved):
I understand that you're repeating the function because we don't have a good "persistent procedure" but this would be a good case for 
RUN isTaxable IN hFinProcs (input x, input y, input z, output a).
ASSIGN taxable = a.

PR #3408, opened 2018-10-02
Ticket - 36400 N-K-1 OEPriceHold has no F1 Lookup.

sewa67:
Files - sys/ref/sys-ctrl.i


brad-advantzware (changes requested):
@sewa67 
OEPriceHold allows for multiple options in a comma separated list. 
F1 should have multi-check interface.  See the pop-up for SSBOLEmail.  The lookup should result
in comma separated char list.

PR #3418, opened 2018-10-03
Ticket - 28730 Make release a pop up box

sewa67:
Files - oe/b-ordrel.w,panels/p-ordrel.w,oe/d-ordrel.w


brad-advantzware (changes requested):
@sewa67  I understand that you don't want to rewrite code that already works but for the code that you copied and pasted, there must be a removal of the code from the source (assume the browse).  We CANNOT HAVE THE SAME CODE IN TWO PLACES, ESPECIALLY IF THE CODE IS NO LONGER CALLED!

brad-advantzware (approved):

PR #3419, opened 2018-10-03
33294 VR11 AP Past Due report needs formatting adjusted

tyndmar:


brad-advantzware (approved):
Change approved but see ticket for additional commentary and suggestion/need for clarification

PR #3427, opened 2018-10-04
35249 BOL Mismatch New Examples

wade-advantzware:
Small addition to debug log


brad-advantzware (approved):
@wade-advantzware won't {1}.{2} always just be report.key-03?  Approving but don't understand it

PR #3437, opened 2018-10-05
Feature/35803 discrepancy with xml and factory ticket

wade-advantzware:


brad-advantzware (approved):
Approving but I don't understand why item # is coming from estimate and not the job.

PR #3440, opened 2018-10-05
#25307 OF1 - Ship To Notes

jgill-advantzware:


brad-advantzware (changes requested):
Check that the value of the array element is not blank before concatenating it into a string.
Only Run the UpdateNoteShip... if the ship_note is not blank

PR #3448, opened 2018-10-08
#25668 Reftable - PREPCADFILE

jgill-advantzware:


brad-advantzware (changes requested):
Do not block comment.  Either run the "Toggle Comment" to comment each line or delete it completely.  For reftable stuff, I would prefer that you delete it completely.  The goal is to run search/grep on reftable and have 0 results.  We can track changes using github

PR #3449, opened 2018-10-08
#25466 Reftable - msf-limit

jgill-advantzware:


brad-advantzware (changes requested):
This removes line 444 from original code:  
{methods/viewers/create/truck.i}
Why?

PR #3466, opened 2018-10-10
Ticket - 36979 Pass # does not appear in Prep/Route Routing table or on pop-up viewer.

sewa67:
Files - ce/b-estop.w, ce/d-estop.w


brad-advantzware (changes requested):
@sewa67 Why was the code to increment the pass removed (for each xop)?  Does the code already auto-increment?

brad-advantzware (approved):

PR #3467, opened 2018-10-10
Ticket - 36955 Convert Order Misc Items to a Pop Up form

sewa67:
Files -oe/b-ordm.w,panels/p-ordmis.w,oe/d-ordm.w


brad-advantzware (changes requested):
@sewa67  - What is the purpose of code in the pAddRecord that counts oe-ordm records?  If you need to know if a record was added, have the d- file return an output parameter.

brad-advantzware (approved):

PR #3475, opened 2018-10-10
34799 Logout procedure should disconnect from databas

wade-advantzware:


brad-advantzware (changes requested):
Do not think we can randomly pull a NK1 value across companies.  Asking Clay for suggested path forward for this

brad-advantzware (approved):

PR #3479, opened 2018-10-11
Ticket - 37053 credit memos.

sewa67:
Files - ar/rep/crdbmemo.i, ar/rep/crdbmemo.p


brad-advantzware (approved):
@sewa67 Approving but in the future, would prefer you remove the code that no longer applies rather than comment.  If you want to comment, use the Toggle Comment to comment each line rather than a comment block

PR #3480, opened 2018-10-11
36416 Financial Statements missing Print Character lines

tyndmar:
revert changes made for 33485


brad-advantzware (approved):
Confirmed with Mark that this is reversion to 16.7.16 program

PR #3486, opened 2018-10-12
Ticket - 34661 <New>NK1-JobCardF=CCC-Hybrid - New ticket for Sheeted Printed Literature Jobs.

sewa67:
Files - cecrep/jobtickL.p, cerep/jobtick.p


brad-advantzware (changes requested):
We can't put this check in anything that isn't specific to 3C.  Some people might have a 303 machine.

PR #3487, opened 2018-10-12
#25307 OF1 - Ship To Notes

jgill-advantzware:


brad-advantzware (changes requested):
The "Footer" procedure.  Where did that come from?  This looks like it was copied from somewhere so now we have 2 instances of it.  A procedure is the right place for it but I don't see where it was removed.

PR #3493, opened 2018-10-12
Feature/#25307 of1   ship to notes

jgill-advantzware:


brad-advantzware (approved):
@jgill-advantzware Code being repeated is never a good thing.  I will approve so that it can be tested but I would like to hear ideas for how you could do this better.  I'm also not certain that replacing carriage returns with "" is going to look OK but I will leave that up to Clay

PR #3505, opened 2018-10-15
Feature/28565 export orders from asi to cti

wade-advantzware:


brad-advantzware (changes requested):
Wade - "CitDir" is used in NK1 lookup.  CTIDir is what is created.

brad-advantzware (approved):

PR #3509, opened 2018-10-16
Ticket - 28257 Request - Prevent Changes in Contract Pricing

sewa67:
Files - est/globquot.w,fg/cust-exp.w,oe/globpric.w,util/ImportCust.p,windows/cust.w,viewerid/custcnt.w


brad-advantzware (changes requested):
@sewa67 your queries that test the .imported field should be the same in all cases.  Since a price matrix may not have a customer, adding FIND cust to the query will eliminate all records with a blank customer.  You can do:
    FIND FIRST cust  NO-LOCK 
      WHERE cust.company EQ cocode 
      AND cust.cust-no EQ  quotehd.cust-no 
      NO-ERROR .
    IF AVAIL cust AND cust.imported EQ YES AND NOT td_imported THEN NEXT.

(I don't think you need the MAIN block reference

brad-advantzware (approved):

PR #3538, opened 2018-10-22
Ticket - 35437 Copy Machine, Edit and Save -> Standards Tab showing previous machine

sewa67:
Files - viewers/mach.w


brad-advantzware (changes requested):
@sewa67 The only change I see is the Obsolete changed to Inactive.  The main defect as originally reported is still there.

brad-advantzware (approved):

PR #3573, opened 2018-11-01
Ticket - 37364 JL8 Enhancements

sewa67:
Files - jcrep/r-jcstdN.w


brad-advantzware (changes requested):
You have a calculation for v-sqft and a calculation for dJobQtyMSF.  Why do you have 2 different ways to calculate MSF?  Use the one that comes from the GetFGArea procedure.
Also, you need to consistently name your variables.  v-sqft is not the naming convention.  dJobQtyMSF is a better variable name and the one that  you should use.  I don't understand why you switch between variable naming styles.  Stick to camel case and start with the data type.
Your indentations do not make sense.  Run the "Correct Indentation". "Correct Case", and "Expand Keywords" option in PDSOE before checking this in.
I also see a "%" in the format for job-qty-msf-fg, that is not correct.

brad-advantzware (approved):

PR #3579, opened 2018-11-02
38276 Items appearing in search that are not on the menu

ron-advantzware:


brad-advantzware (commented):
@ron-advantzware  I will assume the requirement from Jay Farr is to only search through menu items that are accessible to the user.  Shouldn't the search just be through xUserMenu?

brad-advantzware (approved):

PR #3581, opened 2018-11-02
36908 RE: addon programs

tyndmar:
commit 1 - remove EXACT duplicates


brad-advantzware (approved):
Approved with conditions:
The changes listed here that remove the "create-on-add" property assignment are not included in the merge (or validated with Ron that this change should be made)
All 132 files that are explicitly referenced will remain in the addon\ folder.
A new ticket is created to resolve these 132 files to move them out of the addon\ folder (assign to Sewa).

PR #3584, opened 2018-11-03
Ticket - 37939 J-R-10 Report will print Board Qty Received Only after Board is Issued.

sewa67:
Files - jcrep/r-bckmchN.w


brad-advantzware (changes requested):
@sewa67 We lost the uom conversion for this.  If the report is to use v-mat-qty as "EA" you need to convert the item's consumption UOM into EA before adding it to total.

brad-advantzware (approved):

PR #3592, opened 2018-11-06
Ticket - 38373 Blank Report

sewa67:
Files - oe/rep/schdrelN.i


brad-advantzware (changes requested):
Sewa Singh Logic required:
If Qty OH < Order Qty selected, skip all records with Qty OH GE Order Qty (remove the OR qty-on hand eq 0)
This option should show orders with Qty OH eq 0 or less than order quantity
If Qty OH > Order Qty selected, skip all records with Qty OH LE Order Qty.  This

brad-advantzware (approved):
This logic does not make sense but I will give Sewa the benefit of the doubt that he tested this since he did not do what I asked.

PR #3618, opened 2018-11-11
38423 - Corrugate Estimate Layout BOM - calculation of shrinkage

brad-advantzware:


tyndmar (changes requested):
If shrinkage is input as 100% this will throw a fatal error (divide by zero)

PR #3623, opened 2018-11-12
Ticket - 37185 H-B Sales Report add Item Name which is the 1st Description Field for the Finished Goods Item.

sewa67:
Files - salrep/r-detjnlN.w


brad-advantzware (changes requested):
See inline comment.  If avail itemfg before using ar-invl buffer is incorrect.

brad-advantzware (approved):

PR #3646, opened 2018-11-14
38962 Error message - "CopyShipNote" not found message in OU1 for ord…

brad-advantzware:
…er 210782


tyndmar (approved):
It *APPEARS* that this change removes all update-type functionality from the browse, presumably in favor of the new pop-up viewer.  Merging for test, but able to revert if necessary

PR #3656, opened 2018-11-16
Ticket - 39041 AR5 to Excel is creating new column when Name includesa Comma.

sewa67:
Files - ar/ar-agingN.i


brad-advantzware (approved):
Why is this export code repeated so many times?  Approving since this should all be replaced by Jasper AOA

PR #3657, opened 2018-11-16
Ticket - 28059 Include Excel .csv in attachment to email.

sewa67:
Files - jcrep/r-ticket.w


brad-advantzware (approved):
Approving but use of .i for these (there are like 20 different variations) really makes this a mess.

PR #3663, opened 2018-11-16
38805 Auto Log Off Monitor

wade-advantzware:


brad-advantzware (changes requested):
Discussed with Wade.  Changes required to look at full _connection table

PR #3667, opened 2018-11-19
Ticket - 38579 IR12 Error - The month must be between 1 and 12 (Error)

sewa67:
Files - fgrep/r-ageinA.w,fgrep/r-ageinN.w


brad-advantzware (changes requested):
Only addresses Spec #2 on ticket.  Needs to look at other errors, including that which was originally reported.

brad-advantzware (changes requested):
It seems like you use the DATE function outside of all calls to fGetDate.  Why not have fGetDate return a DATE type?

brad-advantzware (changes requested):
There is a display statement in the fGetDate procedure

brad-advantzware (approved):

PR #3688, opened 2018-11-22
Ticket - 39221 Export from MF1 - Should not fill in values if MF1 View Item screen does not display the fields.

sewa67:
Files - rm/dRMExport.w


brad-advantzware (changes requested):
@sewa67 Instead of listing the material types in comma separated lists over and over again, create a variable/constant that houses each comma separated list and refer to the variable in your case/conditions.

brad-advantzware (approved):

PR #3690, opened 2018-11-22
39316 When you go to Item, Total Tab, Press Update, Update estimate - Change Options

sewa67:
Files - cec/kstyle.i,sys/inc/kstyle.i,viewers/itemfgt.w


brad-advantzware (changes requested):
Spec changes requested through ticket.  This change does not meet spec.  Have a question as to what the removal of i = i = 1 does.

PR #3692, opened 2018-11-23
Ticket - 39327 Sharp Shooter Transfer Button V 16.8.8 - Autofind by Job

sewa67:
please review this change and merge into development branch 
Files - addon/windows/l-ldtag3.w


brad-advantzware (changes requested):
Sewa Singh Can you explain why your change will prevent the hanging on search by Job#?  It looks like  you are just building a temp-table with each search selection.
Why not build the temp-table once, and set indexes on the temp-table for the 3 fields that it can search on?

brad-advantzware (approved):

PR #3696, opened 2018-11-27
32294 Remove Release Error Log - Use audit table instead

ron-advantzware:


brad-advantzware (approved):
@ron-advantzware Good stuff. Would it make sense to replace all "CREATE Audit..." calls in code to use this super proc.  If you agree, please enter a ticket to refactor that.

PR #3709, opened 2018-11-30
40027 Unable to upgrade to v16.8.4

tyndmar:
replicates an override done for Valley at 16.7.20 which was not merged into development


brad-advantzware (approved):
This is OK, although a wrapper function would be preferred.

PR #3711, opened 2018-11-30
Ticket - 37371 Custom AP Check Format quote needed.

sewa67:
Files - ap/r-prnchk.w, sys/ref/sys-ctrl.i, ap/ap-ckapfmt2.p


brad-advantzware (changes requested):
Subject to close, check should be in Xprint format

brad-advantzware (changes requested):
Your block of code where you place the MICR code is not correct.  With Xprint, the font is set with the last <F tag it encounters>
So it should be 
PUT "<FMICR Encoding><P20>"
      "<C15>" cBankCode FORMAT "x(30)"
     "<FCourier New><P12>"

brad-advantzware (approved):

PR #3726, opened 2018-12-03
Ticket - 39253 Pick Ticket Notes.

sewa67:
Files - oe/rep/relpremx.i


brad-advantzware (changes requested):
Sewa - the file NotesProcs has procedures that could help here.  See GetNotesArrayForObject to see if might be what you need.  This should output an array of notes, parsed on a certain character size, and tells you how many lines there are, in total.  If that doesn't work, please suggest a change to me to accomodate the case for this ticket.

brad-advantzware (approved):

PR #3729, opened 2018-12-03
Ticket - 38446 Print purchase orders - will not print in version 16.7.20

sewa67:
Files - porep/r-poprt.w


brad-advantzware (changes requested):
Sewa - For each should just be a find first, matching directly on v-start-po (since the condition is that they must be same anyway)

brad-advantzware (approved):

PR #3739, opened 2018-12-04
Ticket - 36022 Request - Exclude Customer from Price Hold Logic

sewa67:
Files - oe/PriceProcs.p

### DIFF
--- a/Legacy/oe/PriceProcs.p
+++ b/Legacy/oe/PriceProcs.p
@@ -67,7 +67,7 @@ PROCEDURE CheckPriceHold:
      Purpose: Checks Price Hold for passed criteria.  Adds record to ttPriceHold table.
      Notes:
     ------------------------------------------------------------------------------*/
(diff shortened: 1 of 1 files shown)


brad-advantzware (changes requested):
Sewa Singh A few things. 
You ignored my instruction on this to disregard the logical value since it is redundant.  User must provide at least one criteria in char value to activate.  If no char value exists, it is not activated.

I also said to add parameters for custID and shipID. You changed the input to rowid.  When you change the parameter to a row id, you can only call a particular procedure or function from within the context of a complete record that must be available to work.  This may make sense for a function or procedure that is performing an action specifically on the record.  Passing specific values empowers the function to be called from anywhere, as long as those value can be passed in.  Besides, the main function already has a cust and ship id parameter.

Callers to CheckPriceHold were not changed or included in this check in.  That will be run time parameter type mismatch.  However, if you follow my instruction, you may not need to change the calling programs since the parameters are already part of the CheckPriceHold.  You should, however, review the caller programs anyway to check to see if those values are passed properly if they are available at the time the procedure is called.

PR #3753, opened 2018-12-07
40525 Main Menu - Add HIghlight to selected Menu Option

ron-advantzware:


brad-advantzware (approved):
@ron-advantzware Just confirming:  there is code in here that doesn't seem related to the change.  It looks like you have changed the "fit in frame" logic.  Merging since I assume that is intended, but you should add comment to ticket if testing should be done for that change

PR #3803, opened 2018-12-15
Ticket - 40651 Calculation on our Job Ticket

sewa67:
Files - cecrep/jobptree.p


brad-advantzware (approved):
Calc seems unnecessarily complicated but not familiar with existing display.  Will approve for testing

PR #3807, opened 2018-12-17
Ticket - 41105 MR3 Raw Material Inventory Report does not look to correct Line item of PO for Sheet Size

sewa67:
Files - rmrep/r-ibtagN.w


brad-advantzware (changes requested):
Sewa - Your change to save the po line and find the po-ordl based on the po-line should be the only change needed.
Your other code (aside from having a developer message "hello" left in there) does not make sense to me.  The sheet dimensions can still come from the job-mat and the cJobNo should be the appropriate job# for the po-line, with that change.  Remove all of the code related to the "OEDATECHANGE" NK1 and remove the dimension conversion function.

brad-advantzware (approved):

PR #3808, opened 2018-12-17
Ticket - 40806 SS Physical Count Raw Material

sewa67:
Files - addon/fg/b-physs.w,addon/rm/b-phys.w,addon/rm/w-phycnt.w,fg/b-phys.w,rm/b-phys.w


brad-advantzware (approved):
Could be common function, but acceptable.  Assuming the query index was pulled from another example or properly researched.  Requires test for slowness.

PR #3817, opened 2018-12-18
Feature/39245 user mode does not save

tyndmar:


brad-advantzware (approved):
Pull request includes .r files.

PR #3833, opened 2018-12-19
41297 Module.r program at protogon had security level set at 900 rath…

tyndmar:
…er than 1000

file changed: asiUpdateENV.w
added module security for Help Maintenance and Import Master
added prgm security for Help Maintenance


brad-advantzware (approved):
@mark-advantzware Approving and this isn't new but what is the reason for the &Scoped-Define that specifies &tablename?  First thought is that this will not show up if we try to look for prgrms.can_run in standard grep.

PR #3844, opened 2018-12-21
Ticket - 41097 Trying to reprint a load tag at Peachtree and we get ablank screen printed - Needs to provide message if file is empty

sewa67:
Files - addon/rm/b-jobret.w,rmrep/rmloadtg.w,rmrep/rmloadtg4.w,rmrep/rmloadtg3.w


brad-advantzware (approved):
@sewa67 Approving but your variable names need to be better. ip-param is a horrible variable/parameter name.  Your name should describe what the value is used for:  eg.  iplReprint  (this is a logical parameter name)

PR #3854, opened 2018-12-24
Ticket - 41195 N-Y-8 Must Work for Estimated Materials

sewa67:
Files - util/rm-mkbin.i


brad-advantzware (changes requested):
Sewa - The main loop requires the .i-code to be "R".  Won't this still filter out estimated material?  Can we have a solution that doesn't use this item.spare-char-2 field?  This seems like it should be a parameter added to rm/rm-mkbin.p to include estimated material.  Additionally, I don't see where the UI is changed to enable the new check box.

brad-advantzware (approved):

PR #3875, opened 2019-01-03
42144 Change to EDI Module to create new shared variables on login

wade-advantzware:


brad-advantzware (changes requested):
@wade-advantzware I don't see how this resolves the shared variable issue.  Can you explain?

brad-advantzware (approved):

PR #3877, opened 2019-01-04
41926 - Fix to Main Menu

tyndmar:
Program Changed: system/mainMenu.w
Improved version integer calculation

Brad, made this change to conform with same function in asiLogin and asiUpdate.
Also, in the function fIntVer , one math statement is faster than the loops with EXP and (at least for me), much easier to read


brad-advantzware (changes requested):
@mark-advantzware The second entry in the version multiplies by 10.  I don't think this is correct.  I think only the 3rd spot should do the multiple by 10.  The other fields should multiply in the final step.
A version with 16.8.5 will be less than 16.8.51 but 16.10.1 will be less than 16.8.5.  My change accounted for that.

PR #3879, opened 2019-01-07
Ticket - 42101 Previous Order# Field not able to be updated that is needed for their Job Ticket

sewa67:
Files - oe/v-ord.w

### DIFF
--- a/Legacy/oe/v-ord.w
+++ b/Legacy/oe/v-ord.w
@@ -3918,8 +3918,8 @@ IF AVAIL xest THEN DO:
       END.
 
       IF xest.ord-no NE 0 THEN fi_prev_order:screen-value = STRING(xest.ord-no).
-      IF oe-ord.TYPE = "T" AND oe-ord.pord- GT 0 THEN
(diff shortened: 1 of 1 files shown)


brad-advantzware (changes requested):
@sewa67 Need an explanation as to what logic you are using to assign this fi_prev_order screen value.  Why does the oe-ord.type = "T" matter?  Where did you get this logic?  Was it there in previous versions?
Also, one assignment only uses "oe-ord.pord-" as a field name.  This may still work and compile but this is not OK to commit.

brad-advantzware (changes requested):
Note the incomplete field names oe-ord.pord-.  Fix that even though this will likely work.

      IF xest.ord-no NE 0 THEN fi_prev_order:screen-value = STRING(xest.ord-no).
      IF oe-ord.TYPE = "T" AND oe-ord.pord- GT 0 THEN
          fi_prev_order:SCREEN-VALUE = STRING(oe-ord.pord-).

brad-advantzware (approved):

PR #3882, opened 2019-01-07
41926 Remove deprecated references to asiLoad

tyndmar:
Programs changed:
system/mainMenu.w
envAdmin/asiUpdate.w
envAdmin/asiUpdateDB.w
envAdmin/asiUpdateEnv.w

changed version testing procedures to conform to standards


brad-advantzware (changes requested):
@mark-advantzware  I really think this should be one program/function across all of the programs.  I didn't know this was in multiple places when I originally wrote the function. 
I also question the 4th spot and the "less than 10" functionality.  If we start to add a 4th element, it should be 2 digits.  0.01 to 0.99.  We shouldn't repeat the mistake of the weird .5 and .51 situation with the potential future 4th level of the version.

PR #3884, opened 2019-01-07
42364 Add ability to launch update from update notification icon insi…

tyndmar:
…de ASI

Program changed: system/mainMenu.w
Changed launch target when new version available and user clicks icon


brad-advantzware (changes requested):
@mark-advantzware  I don't understand this.  Why did we switch from the NK1 to hard-coding the path?  If we do this, we need to change the existing NK1.  But then the code still requires the description to be non-blank?

brad-advantzware (changes requested):
@mark-advantzware sysconst.i  - changed to 16.8.4?

PR #3892, opened 2019-01-10
Ticket - 40794 Request - Change PDF Filename of QUOPRINT=Premier-Excel with Print to Email

sewa67:
Files - est/r-quoprt.w


brad-advantzware (approved):
@sewa67 I will approve this but coding wise, this is not very clean.  the lv-pdf-file could be the only thing changed conditionally rather than the whole IF tb_Hide... block.  Even better, why are the pdf file names not stored and recalled the same for all formats (that is architectural so less expecting of that kind of improvement).

PR #3894, opened 2019-01-10
40531 FG Receipts from Sharp Shooter not updating GL

wade-advantzware:


brad-advantzware (commented):
@wade-advantzware  Why are the following needed to be copied and pasted into this program? I assume these are called *within* the fg-post.i which is obnoxiously bad (not blaming you).  Can we do something better?:
all the variables (not properly formatted)
calc-partial (not properly formatted)
calc-total (not properly formatted)
orig (not properly formatted)

brad-advantzware (approved):

PR #3896, opened 2019-01-11
Ticket - 42164 Bank Reconcilation to allow filter by bank number

sewa67:
Files - ap/b-reconc.w,ap/w-reconl.w


brad-advantzware (approved):
Lots of UI Change.  Approving to be tested.
reconcile temp-table has no indexes (But it never did) so may not be issue with small data set.  Need to test efficiency with large (but realistic) data set.

PR #3897, opened 2019-01-11
Ticket - 40528 Import Error - AF1

sewa67:
Files - util/Importer.w


brad-advantzware (changes requested):
@sewa67 This check is likely too late in the process.  From what I can see in the screen shot, the problem appears to be in READING the records from the csv file.  This would only protect after the records have already been read.  need to add this counter further upstream as ttImportData records are being created.  Per spec, the ttImportData should only have 1000 records in it.

brad-advantzware (changes requested):
@sewa67 This really isn't different than last change.   The system needs to prevent the creation of ttImportData if the record count reaches 1000.  This is a logic change and should be in the pLoad program of ImportProcs.p.  Consider a counter within there and consider adding a return value that can be tested to communicate to the user within Importer.w.  Assign to me if you are not sure

brad-advantzware (approved):
Approving to get code merged.  I will clean up before testing

PR #3936, opened 2019-01-18
Ticket - 42946 1099 Report

sewa67:
Files - aprep/r-1099m.w


jay-advantzware (commented):
It should only do this if run for checks - if run by invoice this is needed - cannot tell from phone if that is done or not

brad-advantzware (approved):

PR #3940, opened 2019-01-18
Feature/34626 create a script to start monitors

wade-advantzware:


brad-advantzware (changes requested):
@wade-advantzware and @mark-advantzware need to work together on this.  I see code that is stale Function IntVer and also code repeated in asiLoginProc.p and ...Bat.p that should only be in one or the other.  Also, session.p should likely house most of the global variables needed at the session level.

PR #3952, opened 2019-01-22
Ticket - 24871 Transfers SHOULD NOT Close Orders See Order 211114

sewa67:
Files - fg/fg-post.i


brad-advantzware (changes requested):
Sent note through ticket.  Want Sewa to try to use CloseOrder.p instead of this call to clsorditm

PR #3961, opened 2019-01-24
33556 User password change enhancements

tyndmar:
File Changed: /patchTemplate/Deployment/Admin/EnvAdmin/prerun.p
added message when password is due to expire in less than 6 days that user should change password using NU3.


brad-advantzware (changes requested):
@mark-advantzware Need to remove the file from the old location

PR #3963, opened 2019-01-24
39730 Estimating record lock when two people are in the same estimate

tyndmar:
File changed: b-estq.w
located procedure which created reftable with buffer
released record after creation


brad-advantzware (commented):
Need explanation of     FIND CURRENT recalc-mr NO-LOCK NO-ERROR.

brad-advantzware (approved):

PR #3964, opened 2019-01-24
40656 Request - Check Box to Declare National Customer

tyndmar:
File changed: viewers/cust.w
added new toggle box, rearranged per spec image


brad-advantzware (changes requested):
@mark-advantzware Should include the schema changes, no?

PR #3971, opened 2019-01-25
Ticket - 43291 Update RELPRINT 10 standard form to print the Loadtag from FG Bins/Jobs tab

sewa67:
Please review this change and merge into development  branch 
Files - oe/rep/relxprnt10.i


brad-advantzware (changes requested):
Determining it is "standard format" requires parsing the first 15 characters, trimming it and then looking for a valid itemfg.i-no that matches.
Your approach will consider anything with more than 15 characters "standard" and that will not work.

brad-advantzware (approved):
Merging to fix myself

PR #4000, opened 2019-01-30
42878 Update repository to better manage deployment files - part 2

tyndmar:
directory restructure and file changes to support same


brad-advantzware (approved):
Approving but need to revisit the need to carry all of these files.  Glad to see Deprecated removed.

PR #4011, opened 2019-01-31
43944 Create delta .df files for DB changes at release time

tyndmar:
Created deltas for asi and aud DBs; using current Devel and 168 and 167 versions


brad-advantzware (commented):
@mark-advantzware OK but isn't there also a master asi.df and audit.df for the db as a whole?  This should only show the changes for this version

PR #4032, opened 2019-02-05
Ticket - 34382 Unipak - Quote needed for updating Pack Out instructions on all estimates with the same CAD

sewa67:
Files - ce/v-est3.w,sys/inc/create_nk1.p,sys/ref/sys-ctrl.i,est/updestpk.p


brad-advantzware (changes requested):
@sewa67 We will change the spec.  The NK1 will be CEPackUpdate.  Additionally, the CEPackUpdate should be a multi-select (like OEDateChange).  We will clean that up in the spec.
You do not need to create a *new* program for this.  You should move the NK1 fetch into updest3.p to avoid the need to pass an argument to it.  Additionally, review the code in the write.trg for eb that assigns values.  This should be moved into this procedure and out of the write.trg.  Please review and confirm that this can be done.

brad-advantzware (approved):

PR #4034, opened 2019-02-05
Ticket - 42328 SS Mod / Feature

sewa67:
Files - addon/fg/b-rcptd.w,addon/rm/b-rcptd.w,addon/b-rcptd.w,addon/b-rcptd.w


brad-advantzware (changes requested):
Requesting Spec Change

brad-advantzware (changes requested):
"Scan Vendor Tags" is the only application in Sharp Shooter that should have this check.  There is no Vendor tag option in Finished goods (yet) and we should only be doing the lookup within the existing RM receipt applications.  See addon/rm/w-recven (v-recven.w)

PR #4040, opened 2019-02-06
Ticket - 43970 O-U-1 (Export to Excel) Ship To Name Problem

sewa67:
Files - oerep/rd-order.w


brad-advantzware (changes requested):
Why is the column "Pallet Count" changed to "Pallet" as part of this ticket?

brad-advantzware (approved):

PR #4043, opened 2019-02-06
44307 ASI user showing in list of users in gear icon

ron-advantzware:


brad-advantzware (changes requested):
NO HARDCODING OF USER ID.
All tests of security level should be centralized in PgmMstrSecur

brad-advantzware (approved):

PR #4045, opened 2019-02-07
Ticket - 43153 Enhancements to 1099 for tax year 2019

sewa67:
Files - viewers/vend.w,windows/l-codedscr.w,aprep/r-1099m.i,aprep/r-1099m.w,aprep/r-v1099A.w,
             aprep/r-v1099N.w


brad-advantzware (approved):

brad-advantzware (changes requested):
@sewa67 The list of 1,2,3 etc, should be centralized so that the # and value only exists in one place in code.  All items can be a LOOKUP function and all uses will be consistent.  Suggest comma separated variables for gc1099Code and gc1099CodeDesc.  Simplest is an include file

brad-advantzware (approved):

PR #4048, opened 2019-02-07
43892 Delete FG Receipts Utility

wade-advantzware:


brad-advantzware (changes requested):
Function code should be separated from UI.  Is there a risk of a transaction being held while this window is open?  What if it is run persistent? We can start a FGTransProcs and move the delete function to that.  UI should just be a shell.

PR #4049, opened 2019-02-07
44497 Inks are missing after upgrade

tyndmar:
Program changed: refTableMigration.cls
Program refTableMigration.cls had incomplete conversion logic for this reftable.
1 - Did not respect extents on reftable.val
2 - Does not throw compile time error
3 - Run time errors are buried in big conversion output table
4 - Did not look up required SECOND reftable to populate eb.side and eb.unit values
5 - Did not create backup reftable1 entries
6 - (Fortunately) did not delete original reftables.


brad-advantzware (changes requested):
.code vs. code2 on Form-no match
substring the .dscr of reftables rather than assign the whole thing
delete reftable (not reftable1)

PR #4067, opened 2019-02-12
Ticket - 44506 Add the ability for a user to export the results of the changes

sewa67:
Files - util/fxtrncst.w


brad-advantzware (changes requested):
The spec calls for an option to output the proposed changes only.  So Add a check box:  "Report on proposed changes only".
When checked, this will not process new assignments of the costs retrieved, it will just show what will happen if they uncheck that option and run the utility again.

brad-advantzware (approved):

PR #4068, opened 2019-02-12
Ticket - 44036 Add Customer Lot# (from O-U-1 Release Tab by the Release# on BOL) to Report

sewa67:
Files -salrep/r-prfinN.i,salrep/r-prfinN.w


brad-advantzware (changes requested):
@sewa67 You do not need to find the .lot-no from the ou1 release tab.  Use the ar-invl.lot-no.  This should be propagated through the order process.
Also, the fix for salrep/getoeret.p parameters is done in another ticket.  We will merge that.  Pull down latest devel since there may be a merge conflict.

brad-advantzware (approved):

PR #4071, opened 2019-02-12
Feature/39060 develop maintenance programs for table ftpconfig

wade-advantzware:


brad-advantzware (changes requested):
Wade - where are the changes within code to use this new ftp table instead of the config files?  We will also need data conversion too, correct?

PR #4099, opened 2019-02-15
TIcket: 42178 Option to Post by User Created and Last Updated User

sewa67:
Ticket: 42178 Option to Post by User Created and Last Updated User
Files:
           fg/fgpstall.w
           fg/FgPostReport.p


brad-advantzware (changes requested):
Request was to "ADD" the parameter filter, not replace Last Updated

brad-advantzware (approved):

PR #4103, opened 2019-02-15
44982 Update DataDigger to latest version

tyndmar:
Downloaded files from DD GitHub
Imported into ASI Repository
See https://github.com/patrickTingen/DataDigger/releases for change history


brad-advantzware (changes requested):
Mark - Why are there .xml files, .dbconnection and .gitignore committed for this?

brad-advantzware (approved):
Approved with the need to commit the deletions of these .xml files

PR #4105, opened 2019-02-16
Ticket - 43765 Is there an NK1 to have SS Job Returns Post Immediately?

sewa67:
Files - addon/rm/b-jobret.w,addon/rm/w-jobret.w,rm/r-rmte&p.w


brad-advantzware (changes requested):
Don't need all the changes to the rm post program

PR #4127, opened 2019-02-22
Ticket - 42525 QUOPRINT = quoprint 1 - Quote Form Modification

sewa67:
Files - cec/quote/quoxprnt.p


brad-advantzware (changes requested):
Need to use NK1 Lookup procedure for "L" value.  You can pass it the context of cust/shipto.

brad-advantzware (approved):

PR #4132, opened 2019-02-23
Ticket - 40816 AF1 Import Flaw

sewa67:
Files - util/ImportCust.p,util/ImportProcs.i


brad-advantzware (changes requested):
Avoid using numbers in your procedure or variable names.  This doesn't help explain why this is different than the original pAssignValueDate proc.
Use something like pAssignValueCToDt, indicating that this assigns using the ignoreblanks logic and converts a character into a Date.

brad-advantzware (approved):

PR #4140, opened 2019-02-25
45140 O-U-1 MISC Tab not Updating TOTALS Tab thus Not Put on Hold

wade-advantzware:


brad-advantzware (changes requested):
Why the sharelock query on oe-ord?

brad-advantzware (approved):

PR #4202, opened 2019-03-08
Ticket: 45395 Order Entry | Export toolbar option - new columns added…

sewa67:
Ticket: 45395 Order Entry | Export toolbar option - new columns added - additional changes requested
oerep/rd-order.w


brad-advantzware (changes requested):
1. Add 'Qty On Hand' as a select-able column
This will be  the  current FG Item OnHand qty (sum Qty for all locations for current company)

You are summing within only one location.  Two options:  use itemfg.q-onh or sum fg-bins.qty for that item.

brad-advantzware (approved):

PR #4204, opened 2019-03-08
46024 Missing Unit Nbrs on InksPack Tab

tyndmar:
Programs changed:
RefTableMigration.cls
Util/dev/reftable-vEst-InkConversion.p
Modified logic to convert ALL unit# reftables, not just ones previously unconverted


brad-advantzware (changes requested):
I don't understand why the original conversion logic wouldn't work.   What was the data case? 
Also, why 2 separate procedures that do the same thing.  If we need to run just that one method on-demand, just call it it in a script rather than repeating basically the whole procedure in a separate file.

PR #4210, opened 2019-03-08
25668 Reftable - [PREPCADFILE]

tyndmar:
Files changed:
RefTableConv.p - added call to method
RefTableMigration.cls - added code to convert reftables to data fields
cerep/r-prepA.w - modified report to display correct fields
cerep/r-prepN.w - modified report to display correct fields
util/dev/reftable-PREPCAD.p - precedure to convert reftable for testing, without upgrade
util/replaceCad.w - modified to populate correct fields
viewers/prep.w - modified to populate correct fields (also updated last-job-no and last-job-no2)


brad-advantzware (changes requested):
There is a big code block removed that did a display or Excel CSV put.  Was that all commented out?
Why do we have the reftable-prepcad.p file?  This is a copy of the conversion.

PR #4213, opened 2019-03-11
Ticket - 36040 Request - Flag for Negative Price Margin

sewa67:
Files - oe/d-oeitem.w


brad-advantzware (changes requested):
Sewa - the .cost field is a cost per M and .t-price is total cost.  oe-ordl.t-cost is calculated total cost.  You need to make sure this check occurs after that calculation though.
Also, the message is supposed to be "Warning: Sell Price is less than the cost"

brad-advantzware (approved):

PR #4221, opened 2019-03-12
Ticket - 45847 M-Q-1 Export to Excel issue

sewa67:
Files - rmiinq/rmiinq-exp.w


brad-advantzware (changes requested):
This will likely solve the problem but you shouldn't present the job number with leading spaces to the user.  Just capture what they enter and add the leading spaces when the report is run.  See R-JobsumN.w for how this is usually done.

PR #4251, opened 2019-03-18
25447 Reftable - ORDERPO

tyndmar:
Files changed:
cecrep/jobptree.p - Removed reftable reference and replaced with job-mat.po-no field
methods/viewers/delete/po-ord.i - Removed reftable reference and replaced with job-mat.po-no field
po/del-po-ordl.p - Removed reftable reference and replaced with job-mat.po-no field
po/doPo.p - Removed reftable reference and replaced with job-mat.po-no field
sys/ref/getpo#.p - Removed reftable reference and replaced with job-mat.po-no field
util/dev/spReftable.p - added proc to convert reftable

### DIFF
--- a/Legacy/cecrep/jobptree.p
+++ b/Legacy/cecrep/jobptree.p
@@ -1366,20 +1366,12 @@ FOR EACH job-mat
       AND INDEX("1234BPR",item.mat-type) GT 0
     NO-LOCK,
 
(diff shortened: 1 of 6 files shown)


brad-advantzware (changes requested):
See Comment in line with code on line 2597 of doPO.p and line 1945 of spReftable.p

brad-advantzware (approved):

PR #4252, opened 2019-03-18
33763 Touchscreen - job thats not on schedule showing up

ron-advantzware:


brad-advantzware (changes requested):
See ticket - New NK1 mentioned in spec

PR #4267, opened 2019-03-20
Feature/38038 warehouse locations and bins inactive flag

wade-advantzware:


brad-advantzware (commented):
@wade-advantzware I thought you were going to move these validations into InventoryProcs.p?  They should probably be called "ValidateLoc" and "ValidateBin".  A "Get" usually returns a calculation or object to use.  If you prefer to do this later, we can merge this but create ticket to move Inventory related validations into InventoryProcs.p

brad-advantzware (changes requested):
Merge Conflict on b-trans.w - large blocks moved?
While fixing that, ValidateBin should be renamed ValidateBinFG

brad-advantzware (approved):

PR #4269, opened 2019-03-20
Feature/32941 improvements to asn for valley

wade-advantzware:


brad-advantzware (changes requested):
Only file location changes suggested - 
Required: Put test files in "testers" folder, not util.  Util should only be for utilities with user access.
(suggested not required: Consider using a folder other than "custom".  "system" may be better)

brad-advantzware (approved):

PR #4273, opened 2019-03-20
33763 Touchscreen - job thats not on schedule showing up v2

ron-advantzware:


brad-advantzware (approved):
@ron-advantzware Approving and merging since a .i is still better than copied code and we have bigger fish to fry.  But...
I would've preferred a "BuildJobs.p" procedure parameterized with the inputs required to build the temp-table (which can be shared).  Get_Jobs.i can only be used within this program due to the reliance on variable referenced in the container (like the new one lTSShowPending).  Someone just looking at Get_Jobs.i is looking at things that aren't explained within that program file.  The reason I'm sensitive to this is that I have been inside the estimating code and this is an absolute code supportability killer (obviously I have seen way worse than this).

PR #4276, opened 2019-03-22
Ticket - 45943 M-R-10 Roll Stock Consumed Report

sewa67:
Files - rmrep/r-consumN.w


brad-advantzware (changes requested):
@sewa67 The procedure rm/convquom.p should be used here to convert between LF and LBs.  You can use the same direct assignment if the UOMs match.

brad-advantzware (approved):

PR #4277, opened 2019-03-22
Ticket - 46525 N-K BOLFMT = SouleMed Page Number is Not Working

sewa67:
Files - oe/rep/bolsolmed.i,oe/rep/bolsolmed.p


brad-advantzware (commented):
@sewa67 If this change was just to fix the defect of page numbering (which looks OK), why is there a large block of code changed, that deals with the w2.cas-cnt and w2.cases (lines 347-462)?

brad-advantzware (approved):
I will approve but this does not seem like it is optimal code.  Not worth reworking a custom form

PR #4279, opened 2019-03-22
Ticket: 46563 N-K InvPrint = SoulePO (Customer Part# is showing FG It…

sewa67:
Ticket: 46563 N-K InvPrint = SoulePO (Customer Part# is showing FG Item #)
ar/rep/invsoulepo.p
         oe/rep/invsoulepo.p


brad-advantzware (approved):
Noted in ticket that this is likely not a complete removal of the sensitivity to the svi=print-item shared var (radio button on invoice param screen)

PR #4281, opened 2019-03-22
46486 Case label creation - calculation of number of labels incorrect

tyndmar:
Files changed:
oerep/d-casetg.w - changed label qty calculation
oerep/r-casetg.w - changed help text on several fields; updated some field positions for cosmetics; moved CheckRelease procedure out of functions area; modified LEAVE of case tag to allow more lookups (don't check for freeze, don't check modified - reports of value-change failures in ticket 46008)
oerep/r-loadtg.w - used the AppBuilder to modify field positions (sorry);  NO LOGIC CHANGES (I know the triggers look screwy in comparison, but they're only rearranged alphabetically)


brad-advantzware (approved):
Note:  Per Mark - no logic changes in r-loadtg.w but need to regression test.

PR #4289, opened 2019-03-25
42966 Add UDF to Factory Ticket

ron-advantzware:


brad-advantzware (commented):
@ron-advantzware Is there no direct identifier for the GRN field inside UDF?  It just looks like you are getting all UDFs for the itemfg.  What if they add another?

PR #4291, opened 2019-03-26
Ticket - 30958 Corr Estimate | Spec: new logic to prompt to update CAD # and Box Image path

sewa67:
Files - cec/v-est.w,est/w-estc.w,custom/d-cadimgupd.w,panels/p-cadimg.w


brad-advantzware (changes requested):
@sewa67 We do not want to add business logic or db changes inside a UI file.  Create a .p layer to the update cad program and use public procedures that can be called from the UI.  The only job of the UI is to capture input from the user.  The code on OK Button needs to be in a procedure.  I would also like you to put the code that queries the estimate data to present form count into a sub proc as well.

PR #4293, opened 2019-03-26
46472 Scanning error to FG item

tyndmar:
Program changed: p-updbar.w
Removed superfluous cancels/adds
Prevented double-save when auto-add active


brad-advantzware (approved):
Can't explain the multiple "APPLY CHOOSE" commands but that was a mess.  Need to include regression test list and recommended tests for this.

PR #4301, opened 2019-03-27
44101 name on quote is not the same as name in Advantzware

tyndmar:
asiUpdate1.w - improved search for advantzware.ini to prevent edge cases for mapped drives and file locations (no ticket involved)
asiUpdateDB1.w - improved search for advantzware.ini to prevent edge cases for mapped drives and file locations
asiUpdateENV1.w - improved search for advantzware.ini to prevent edge cases for mapped drives and file locations
- tkt 44101 - added ability to modify advantzware.pf file - specifically reassign cpstream parameter
-- lines 92-100 and procedure ipUpdateAdvantzwarePf
- no ticket - added insurance that all tables in asi DB are referenced in auditTbl file (Ron's request) - lines 3266-3281
buildDumpFiles.p - ticket 42096 - suppressed export of permissions from AWS db

### DIFF
--- a/Template/BuildScript/ToCompile/asiUpdate1.w
+++ b/Template/BuildScript/ToCompile/asiUpdate1.w
@@ -1004,104 +1004,68 @@ PROCEDURE ipFindIniFile :
       Parameters:  <none>
       Notes:       
     ------------------------------------------------------------------------------*/
+    DEF VAR cLetters AS CHAR NO-UNDO INITIAL "ABCDEFGHIJKLMNOPQRSTUVWXYZ".
+    DEF VAR cEnt AS CHAR NO-UNDO.
+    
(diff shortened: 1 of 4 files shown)


brad-advantzware (changes requested):
@mark-advantzware Why are there 2 copies of the find advantzware.ini logic?  We should do it once and in one program.  Are there other files you have to hunt for like this?  If so, it should be generalized.
On the auto create of Audit files, is the spec to create them all off by default?  I thought we activated Create and delete by default?  Lots of tickets here so not sure where to look.  If no ticket, one should've been created.
Re: export edit, what is this dumping from?  We need to ensure we seed the data properly and the real test in when we *import* the file that we don't touch permissions.

PR #4308, opened 2019-03-28
Ticket - 44303 Unable to Print Load Tag - Ships Tomorrow

sewa67:
Files - oerep/d-loadtg.w


brad-advantzware (approved):
@sewa67 I will merge this but you need to explain why you changed the WINDOW-CLOSE to now run an EMPTY TEMP-TABLE when it did not before.  On the surface, I would've expected that you would just remove the pCheckTag call you added.

PR #4323, opened 2019-03-28
45140 O-U-1 MISC Tab not Updating TOTALS Tab thus Not Put on HOLD

wade-advantzware:


brad-advantzware (approved):
@wade-advantzware I would probably fail this is we weren't at the end of the release and you weren't trying to prep for a visit tomorrow.  
In the future, this should've been an internal procedure.  Same block of code copied in more than one place.

PR #4357, opened 2019-04-02
Ticket - 46943 Need ability to control adjusting of bins in I-F-1

sewa67:
Files - browsers/fgijob.w


brad-advantzware (changes requested):
@sewa67 If "viewers/p-fg-bj-l.w" is no longer in use inside PgmMstrSecur (no other callers) remove it from the case statement within PgmMstrSecur.p

brad-advantzware (approved):

PR #4370, opened 2019-04-04
47367 OR6 REPORT

ron-advantzware:


brad-advantzware (approved):
@ron-advantzware painful to see repeated code but it was already there.  Need to confirm that our output call to make a .CSV line is done ONE WAY only going forward, hence the question about Jasper

PR #4383, opened 2019-04-08
47525 E-F Ink Panel, warnings

tyndmar:
Program changed:
ce/v-est3.w
added EXCLUSIVE option to FOR EACHes that update fields.


brad-advantzware (changes requested):
The block with For each b-ef of eb that was changed from a NO-LOCK to EXCLUSIVE should likely be completely deleted.  This looks like the purpose was for reftable manipulation and since we are working with native db fields, I don't think this serves a purpose (other than to cause the NO-LOCK error).

PR #4406, opened 2019-04-12
47794 E-B-3 Scores Button No Longer Works

wade-advantzware:


brad-advantzware (commented):
@wade-advantzware Is there a reason that we need to manipulate the ERROR-STATUS for this?  I would really like to avoid raising an ERROR condition and having to clear the error.  Would it be possible to just return an output parameter?  It seems like if this just stopped working that we can't rely on consistent behavior of the error status, possibly because of upstream dependencies.

brad-advantzware (changes requested):
Wade Kaldawi As discussed with Ron, the real culprit here is that the timer ctrl code sets the error status to YES on a find statement.  We should fix the root problem in main menu since this situation may arise elsewhere.
Raising for standup to notify others that this may have been a problem and discuss the merits (or lack thereof, in my opinion) of testing ERROR-STATUS with validation.

PR #4422, opened 2019-04-16
Ticket - 46717 N-K-1 JobCardF = Unipak Job Due Date is printing from Order View Tab

sewa67:
Files - cerep/jobunipk.p

### DIFF
--- a/Legacy/cerep/jobunipk.p
+++ b/Legacy/cerep/jobunipk.p
@@ -73,6 +73,7 @@ def TEMP-TABLE w-lo NO-UNDO
   field layout like v-layout.
 
 def new shared buffer xjob-hdr for job-hdr.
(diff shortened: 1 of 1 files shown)


brad-advantzware (changes requested):
Why did the existing job-hrd buffer not provide the correct item# in order to find the appropriate oe-ordl?  This seems like there would be more problems with accessing the wrong job-hdr than just the Due Date for the corresponding order line.  The block of if "v-first" I think needs to be managed better.  Since this oe-ordl find should really be done per job-hdr from the main loop.  If I'm misreading, please explain.

brad-advantzware (approved):
Warned that this may impact printing of Sets with this job ticket but merging for testing.

PR #4446, opened 2019-04-19
46030 Open Windows Not Coming to Front

tyndmar:
Program Changed: nosweat/persist.p
Modified procedure that launches programs, in the section that checks whether the pgram is already open, to find the handle of the running window and bring to the top of the visual stack.


brad-advantzware (approved):
Approving but need to confirm Golden Code support.  Equivalent to User32.dll in linux.

PR #4449, opened 2019-04-19
28308 Release Ticket Numbering (OT1)

tyndmar:
File changed: oe/release#.p
Modified extra release# check to test against oe-relh versus oe-bolh.
Also, implemented spare-char-1 data value on Protagon (was blank) to prevent duplications


brad-advantzware (changes requested):
Move all "extra" checks intot he getNextRelNo.p for all request types.
Do an additional check to make sure the .r-no and release# assignments in the oe-xxxx tables use the getNextRelNo.p.  When processing an unassembled set, I encountered the .r-no already used error.  Let me know if you can't find that and I will try to create steps to reproduce.

PR #4451, opened 2019-04-19
47174 Jasper - Task Monitor 32 bit vs 64 bit

ron-advantzware:


brad-advantzware (changes requested):
Movinto Session procs and consider the method described in: https://knowledgebase.progress.com/articles/Article/000054631

PR #4452, opened 2019-04-19
45721 Utilizing cXML Monitor to Pull TARGET Orders

wade-advantzware:


brad-advantzware (changes requested):
@wade-advantzware We need to key order line creation off of something more concrete.  Suggestion is that we have to key off "ItemOut" and create a new ttOrderLine once that is encountered.
Please review and see if that is a better, more reliable method rather than hoping for or looking for the "last" node.
Additionally, we will need a tester utililty (or something that stops short of importing the order) that outputs the import temp-tables for review.

brad-advantzware (changes requested):
Variable declaration should be moved to top. 
Clean the find first WHERE NO-ERROR (remove WHERE)
Remove unused variable declarations.
Tester is a bit overkill but fine

PR #4467, opened 2019-04-23
Feature/28920 count duplicate message appears when you edit any count entry should only be on add

sewa67:
Files - fg/b-phys.w


brad-advantzware (changes requested):
Sewa.  Change the message that you coded from:

"There is already a count entry for this tag in location '" + bf-fg-rctd.loc + "' with a count of '"
+ STRING(bf-fg-rctd.qty-case) + "' Are you sure? with an OK "

To "There is already a count entry for this tag in location '" + bf-fg-rctd.loc + "' with a quantity of '" <use the fg-rctd's total quantity instead of case count> + ".  Are you sure you want to add another count entry for this tag?".
(the "with an OK" was not part of the message in spec.)

Check other Messages in this program and refer to the "quantity" not the "count" and display the total quantity, not the case count.

brad-advantzware (approved):

PR #4474, opened 2019-04-24
Ticket - 47013 PO on hold receipt via rmtag

sewa67:
Files - rmrep/rmloadtg4.w


brad-advantzware (changes requested):
Do not use ERROR-STATUS to return an error in validation.   Just discussed this on standup yesterday

brad-advantzware (changes requested):
@sewa67 You can return a logical from the pValidatePO procedure or you can make it a function that returns a logical and do If NOT fIsValidPo(lv-po-no) THEN RETURN NO-APPLY.

PR #4489, opened 2019-04-25
Ticket - 46134 Print or Post programs default the name of the file to the program name - running again causes Err (98)

sewa67:
FIles - gl/r-gljrnN.w


brad-advantzware (changes requested):
Sewa Singh You are about to make a change like this in dozens of programs.  Are you going to copy the same code to every program?  What if Jay then says, we want to change the format of the auto-file that is generated.  Are you going to go into those dozens of programs to make the change again to all the places where you put the file name together?

PR #4495, opened 2019-04-25
32941 Improvements to ASN for Valley

wade-advantzware:


brad-advantzware (commented):
Brad Vigrass Private . 07:42 PM

Wade Kaldawi Nice job on this.  I love your encapsulation of these procedures into one procedure library.  I also appreciate the effort to not have any business logic in the .w file, even in the case of a utility. 
One question I have is the DELETE object inside the monitor.  Is that necessary?  Won't that add a bit of overhead to constantly create and delete the program object with each monitor fire?  I'm no expert so just looking for opinion from Mark Tyndall or Ron Stark on that too.

PR #4498, opened 2019-04-26
Ticket - 48158 F1 Lookup on PO for FG not displaying all active items when "All Vendors" is selected

sewa67:
Files - windows/l-itemf2.w,windows/l-itmall.w


brad-advantzware (changes requested):
@sewa67 I don't see a difference between your open-query, other than a different sort-by.  My suggestion of OUTER-JOIN was just a suggestion.  Won't this result in the "PO Vendor" list not being filtered properly and showing everything?  Since you have 2 different query blocks, I would think you don't use an outer-join. Instead for the "All Vendors" case, you don't join the e-item... tables at all and on the PO Vendor case, you join on e-item... tables without an OUTER-JOIN.

PR #4503, opened 2019-04-28
48083 Field Level View of Audit History data for any specific field u…

tyndmar:
…sing a hot key

See Zoho comment for details


brad-advantzware (changes requested):
After discussion, I would really like to avoid introducing another shared var.  I generally agree that we should really be pulling the session company from the Session.p persistent.

brad-advantzware (approved):

PR #4515, opened 2019-04-30
Ticket - 42646 Finished Goods - 'Pallet type' added to FG Item database and Loadtag data file

sewa67:
Files - cec/v-est3.w,oe/d-oeitem.w,oe/v-ord.w,cec/UpdFgEc.i


brad-advantzware (changes requested):
Your check-ins for v-est3 and d-oe-itemw are fine.
I'm failing this because of your use of a .i within a .w that performs a DB transaction to update values from E-C into itemfg.  Here is what I want you do.
Review E-F for the fields in the Inks/Pack tab that are "synched" with their fields in I-F-1.  This includes (but need you to verify that this is all of them):
Case/Unit Count
Cases/Units per Pallet
Pallet # (new)
Zone (new)
The code that updates the cases and pallets is likely legacy code that doesn't seem to cause a problem.
Try to use the *same* approach with updating the itemfg fields from E-C that appears to successfully work for E-F.

PR #4521, opened 2019-04-30
42577 Request - Auto Release of Web Orders

tyndmar:
This is the "guts" of changing order status to Hold if one or more tests fails
/testers/smart files that allow manipulation of the validationTest table - this table is an extensible list of tests that must pass to approve orders
custom/oe-ctrl.w - selects tables to display to a user Admin to turn on/off validation testing
system/oeValidate.p - structured procedure containing the specific test procedures
system/tagProcs.p - modifications to support Hold tags


brad-advantzware (changes requested):
Changes requested documented in ticket

PR #4535, opened 2019-05-02
Ticket - 43596 Request - Bill by PO once entire PO is complete and shipped

sewa67:
Files - oe/b-oeinv.w,oe/invhold.w,oe/v-oeinv.w,viewers/cust.w


brad-advantzware (changes requested):
In v-oeinv.w, you MUST include the Customer PO matching when searching for oe-ordl (in the "fully shipped" test) and when finding other invoice Headers to set to H>  You are only looking at Order # (ord-no) which is not sufficient.  inv-line.po-no is the key variable and must match oe-ordl.po-no similar to how you handled the oe-rel query.

brad-advantzware (approved):

PR #4555, opened 2019-05-06
Ticket - 45621 Request - Add Routing to IR1

sewa67:
Files - fgrep/r-fgrordN.i,fgrep/r-fgrordN.w,fgrep/r-fgrordww.i


brad-advantzware (changes requested):
Sewa - Good job making this a function.  However, check out JobProcs.p.  There is a GetOperations procedure.  Since this one is a little different, I suggest adding a GetOperationsForEst and pass the est-no and .i-no (match on stock-no) to get the routing from est-op.  Need to get the eb first to match the form #

PR #4598, opened 2019-05-14
49038 ITEM HALDEX88725/Cost

wade-advantzware:


brad-advantzware (changes requested):
create new GetCostForLastReceipt(company, i-no, [output same as GetCostForReceipt]).
Internally, find the last "R" first, if zero then "C",
Do same conversion.
Apply, detail costs to itemfg. fields as well.

brad-advantzware (approved):

PR #4602, opened 2019-05-15
Ticket49016 Add ability to export and import carrier matrix

sewa67:
Please review the changes and merge into development branch.
Files: windows/carrier.w , browsers/carr-mtx.w , fg/carr-mtx-exp.w , fg/carr-exp.w , util/importCarrier.p , util/ dev/impcarrier.p , util/ttimport.i , util/ImportCarrierMtx.p , util/dev/impCarrierMtx.p.


brad-advantzware (changes requested):
Sewa - A few typos:

    "Key Field Blank: Change Method.". (should be "Charge Method")
    IsValidFromList IN hdValidator ("Charge Method",ipbf-ttImportCarrier.chg-method, "MAF,Pallet,Weight", (should be MSF)

Also, the Column Label should refer to "Quantity" and not Pallets since the Charge method may be Weight or MSF and doesn't have to be pallets.
So change columns to "Qty Up To" and "Rate/Qty"

Also, I don't see the following requirements implemented.  Please refer to the way that Inactive ShipTos are set.  This uses Tag Procs to validate if something is inactive and to add an inactive tag:
4) The carrier will contain an Inactive flag on the carrier file such that the inactive flag prevents the carrier from appearing on the F1 lookup and this likely should be part of the new F1 lookup for carrier such that it is coded once
5) The carrier being inactive should be prevented from being added to a new record that is added with at least these areas verifying the inactive status when adding a new record or updating an existing record:
- Customer File
- Ship to File
- EstRelease
- Estimate
- Order
- Release
- Pick Ticket

brad-advantzware (approved):

PR #4669, opened 2019-05-28
Ticket - 45714 JobCardF=CentBox - Enhancement To Facilitate Sheeting

sewa67:
Files - cerep/jobcbox.p


brad-advantzware (approved):
@sewa67 This is OK but the repeated code in the "If j EQ 1" could be turned into a sub procedure.  The positions of the <R and <C tags could be changed to be dependent on input parameters, but I will approve this since it is a custom form.

PR #4674, opened 2019-05-29
Ticket - 49707 Xprint "phantom box"

sewa67:
Files - oerep/r-bolprt.w


brad-advantzware (changes requested):
Checking with Sewa to see if removal of the <PROGRESS> tag would be better approach rather than the end </Progress> tag

brad-advantzware (approved):

PR #4679, opened 2019-05-29
49855 SB DMI Export Import Job format

ron-advantzware:


brad-advantzware (changes requested):
@ron-advantzware Ron - The job parsing rules need to be centralized.  Is this *the* way we're going to do this?  
See jc/JobProcs.p - JobParser (written by JKTech) - we should have these rules for parsing a job string in one place so that we can ensure we are standardizing our formats for job bar codes.  
Additionally, we should probably create a JobAssembler proc that does the reverse.  Do you agree?

PR #4687, opened 2019-05-31
Ticket - 49774 Label Matrix - "Forced" NK1 BARDIR integer value 1 functionality

sewa67:
Files - oerep/r-loadtg.w


brad-advantzware (changes requested):
Sewa - I would like to centralize this function and it should be done inside OutputProcs.p
Please review the GetBarDirFilePath: procedure.  It may make sense to enhance the GetBarDir.p program too.

brad-advantzware (approved):

PR #4719, opened 2019-06-10
Ticket - 49707 Xprint "phantom box"

sewa67:
Files - system/OutputProcs.p

### DIFF
--- a/Legacy/system/OutputProcs.p
+++ b/Legacy/system/OutputProcs.p
@@ -390,8 +390,7 @@ PROCEDURE WriteToXprintRect:
 
     cText = "<||><R" + STRING(ipdRFrom) + "><C" + STRING(ipdCFrom) +
             "><FROM><R" + STRING(ipdRTo) + "><C" + STRING(ipdCTo) + "><RECT>" . 
-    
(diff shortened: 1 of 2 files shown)


brad-advantzware (changes requested):
Sewa - This is good so far but I think the Font procedure needs an "F" in stead of a "R".  Additionally, please enhance the run-xprint.w utility to include these sample outputs using the OuputProcs.p.  Change the run-xprint.w to not write directly to the output file, instead change it to use the OutputProcs.p so that we can verify.  Add this to the TEST case when submitting the code for review.

brad-advantzware (changes requested):
@sewa67 when you name your variable handles, it should reflect the object that it represents.  In this case you use hftp, likely because this declaration was copied from somewhere else.  The handle should "hOutputProcs".
Additionally, I'm also interested in how OutputProcs will work when doing additional business forms.  The text that you manually have written through Put ("Sold To:" etc.)  change this to use procedure calls to OutputProcs.  Put one rectangle around Each of the Ship To and Sold to blocks of text.  If you would like to enhance things, let me know where you think we could improve.

PR #4723, opened 2019-06-11
Ticket - 48549 I-U-1: Whse and Bin fields not saving non-default value

sewa67:
Files - fg/d-rcptd.w


brad-advantzware (changes requested):
You assign the cPOBeforeChange to the job-no.  This does not seem correct.
You're name for this variable is good, although you did declare it globally so it should be gcPOBeforeChange.
Your naming on the new parameter you added to the pDisplayFG proc should be a logical.  Call it iplGetLocBin.  This provides clearer self-documentation.

PR #4734, opened 2019-06-13
48501 Pending by Job - Add customer column

ron-advantzware:


brad-advantzware (changes requested):
Remove your direct contact information from the about.txt.  Just have "help@advantzware.com"
Not sure that is our new number or not

PR #4741, opened 2019-06-17
Ticket - 50523 Issue: P-Q-1- No receipts

sewa67:
Files - fginq/j-fgiinq.i,oeinq/b-ordfgi.w,poinq/f-porec.w,viewers/d-fg-rcpth.w


brad-advantzware (changes requested):
Sewa - I am concerned with changes to f-porec.w and j-fgiinq.i - 
In the latter, the "vend" index is Use through USE-INDEX.  This *requires* vendor code matching and your removal of that criteria could lead to a very slow query through fg history.
You need to use either the "vend" using both vend-no and po-no or "item-po" which requires "i-no" and "po-no".

brad-advantzware (approved):

PR #4773, opened 2019-06-20
Ticket - 50442 Error creating trailer in AF8

sewa67:
File - viewers/truck.w

### DIFF
--- a/Legacy/viewers/truck.w
+++ b/Legacy/viewers/truck.w
@@ -34,15 +34,10 @@ CREATE WIDGET-POOL.
 /* Parameters Definitions ---                                           */
 
 /* Local Variable Definitions ---                                       */
-{custom/globdefs.i}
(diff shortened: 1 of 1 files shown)


brad-advantzware (changes requested):
Sewa Singh A change to the viewers/truck.w file was made with this commit:
https://github.com/Advantzware/Advantzware/commit/6a59e1a969212c9be4a556a3ff65035940344078#diff-e29f7c150fb26352ca7adfe9d37015ac
Note that the line:
{methods/viewers/create/truck.i} was removed, it appears, in error.
Would the fix for this simply be to restore that line, rather than all of the changes you made.  Please review change logs to programs before coding to look for the simplest possible change to make.  Especially something that appears to have been recently broken.

brad-advantzware (approved):

PR #4777, opened 2019-06-20
49464 Wrong name on BOL

wade-advantzware:


brad-advantzware (changes requested):
I have a concern with this, just because I thought the point is that the enteredBy is being entered TOO often.  This adds even more places where the enteredBy is filled. In Checking with CLay

PR #4794, opened 2019-06-24
Ticket - 47653 Lineal Feet

sewa67:
File - pcrep/r-mcheff.w, pcrep/r-mcheff.i


brad-advantzware (approved):
Approving.  Notified testers that math may not be right in this.  Need to sum the total run qty in LF divided by the total hours for each run.

PR #4801, opened 2019-06-25
50767 Error Message with Query Builder

tyndmar:
moved field selector dialog from util/dev to /util


brad-advantzware (changes requested):
I was expecting a "file was moved/deleted" check in here.  This looks like just an "add"

PR #4804, opened 2019-06-25
50205 Order #915561

wade-advantzware:


brad-advantzware (changes requested):
Wade - If the Job Number is blank, the PRoduction Qty should be 0.  Was there a case where the USE-INDEX i-no would be needed?

brad-advantzware (approved):

PR #4808, opened 2019-06-25
Feature/49636   web orders   auto approve and release

wade-advantzware:


brad-advantzware (changes requested):
Check Public Private and name according.  
Private Procs can pass buffers instead of rowids if that makes sense
All NK1s should be run in a pGetSettings that populates global variables/properties within the procedure/class
Convert all NK1s to NK1Lookup
Add "this proc will replace xxxx" to comments section of procedure for further refactoring
Need to reinstate the balance calculations for itemfg, itemfg-loc

PR #4821, opened 2019-06-27
Ticket - 49854 OEShipFrom and OEShip logical values are confused

sewa67:
Files - oe/d-oeitem.w,sys/inc/create_nk1.p


brad-advantzware (changes requested):
Sewa - use oe-ordl.t-cost.  This will simplify the request and will also still work if should oe-ordl.cost ever becomes something other than per M (possible)

PR #4833, opened 2019-06-28
Ticket - 49651 Update the PO button in I-F-1 to display the Location information

sewa67:
Files - browsers/itemfgpo.w


brad-advantzware (commented):
@sewa67 Why do you need to use cocode here and not just po-ord.company ?  It looks like the buffer is in use for po-ord.vend-no already?  Merging this but need to understand why

PR #4841, opened 2019-06-29
Ticket - 47531 V-T-4 - Bank Reconciliation

sewa67:
Files - ap/b-recons.w
ap/w-reconl.w
util/ttImport.i
ap/ExpBankReconl.w  - new file
util/ImportBankReconl.p - new file
util/dev/impBankReconl.p - new file


brad-advantzware (changes requested):
Need more comments.  ap/reconcil.p - what does this do (need to state that it builds the reconcile tt and what else?...
There is a no-lock on the reconcile find but an assign into it.  That will need to be exclusive-lock.
You should create a buffer parameter (ipbf-reconcile) to reconcile-file and pass the buffer instead of doing an additional find inside the sub procedure.
Qualify all field references to the temp-table buffer.  e.g. ipbf-reconcile.tt-type.
You should turn the IF Then ELSE block into a CASE statement.

PR #4851, opened 2019-07-05
Ticket - 51209 New report for N-K-1 Settings to be selectable column and allow for Help to be printed

sewa67:
Files - windows/sys-ctrl.w,windows/sysctrl-exp.w,viewers/prtsyscl.w,windows/r-sysctl.w


brad-advantzware (changes requested):
The buffer parameter ipb-itemfg FOR sys-ctrl is not acceptable.  
Buffer variables need to reflect the table that it is for  ->  ipbf-sys-ctrl

The function getValue-itemfg is also bad.  This was all clearly copied and pasted with no attention given to having the functions or names representative of what the functions and variables actually do.  Just calling it getValue would be better but something like "fGetValueExternalToSysCtrl" is more appropriate.

PR #4913, opened 2019-07-19
51977 Slow Scan Times via Sharpshooter

wade-advantzware:


brad-advantzware (approved):
I don't completely understand why the first r-no is required to be part of the query.  Assuming it is for index purposes.  SharpShooter rewrite will make this irrelevant so approving

PR #4914, opened 2019-07-19
52019 Que Card Improvements

ron-advantzware:


brad-advantzware (changes requested):
There looks to be duplication of code here.  pInactivateCueCards in MainMenu and spInactivateCueCards in session.p.  Also, does it make sense to have a UI element (message with update) in a persistent .p?  I like encapsulating all functions in one place but the session object seems like a place we don't want UI prompts.  Call if easier.

PR #4960, opened 2019-07-26
52427 DAOA Translate Module to Menu Sub Group to invoke 0 (zero) menu…

ron-advantzware:
… option


brad-advantzware (changes requested):
brad and ron to review if there is a better option to map modules/hotkeys

PR #4962, opened 2019-07-26
Ticket - 48279 Do not allow an inactive ship to used on an estimate or order

sewa67:
Please review this change and merge development branch 
Files - ce/b-estitm.w
ce/b-estitm1.w
cec/b-estitm.w
methods/viewers/assign/cust.i
oe/v-oeinv.w
oe/v-ord.w
viewers/shipto.w
 ce/v-est.w
cec/v-est.w

### DIFF
--- a/Legacy/ce/b-estitm.w
+++ b/Legacy/ce/b-estitm.w
@@ -4793,14 +4793,14 @@ PROCEDURE new-cust-no :
       FIND FIRST shipto
           WHERE shipto.company EQ cust.company
             AND shipto.cust-no EQ cust.cust-no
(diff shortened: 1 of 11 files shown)


brad-advantzware (changes requested):
Sewa Singh You need to remove the changes related to the .isDefault flag in shipTo.  We aren't ready for that and it was not in Jay's spec (his mention of default said "Same ship to code as customer").
I think you are on the right track with those changes, but this ticket is just to validate and block inactive shipTos.
Additionally, you should have one program that validates a shipto, accepting company, customer and shipto as inputs (can you use the same process as Imports? validate.p?)

PR #4976, opened 2019-07-29
51300 Cannot add new ship to from estimate

tyndmar:
Programs changed:
ce/b-estitm.w
ce/b-estitm1.w
cec/b-estitm.w
Added logic in leave trigger of cust-no to find best match of ship-to and populate the ship-id field with this entry

### DIFF
--- a/Legacy/ce/b-estitm.w
+++ b/Legacy/ce/b-estitm.w
@@ -923,6 +923,38 @@ DO:
 
     RUN valid-cust-user NO-ERROR.
     IF ERROR-STATUS:ERROR THEN RETURN NO-APPLY.
+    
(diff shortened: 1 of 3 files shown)


brad-advantzware (changes requested):
Need to put into a procedure.  Suggest:  GetShipToDefault (company, cust-no, output ship-id).
The query that test ship-no eq 1 will never be run.  This should likely run as the 3rd option before the "any shipto for customer" logic.
This could be put into a "system/CustProcs.p" and exposed as a public proc/function.  There could be other procedures that take a customer as input and do something or return something too.  Would be better to not have multiple files for all of that.

PR #4990, opened 2019-07-30
51873 N-K-1 InvPrint = Xprint Invoice Not combining BOL's with same item

tyndmar:
Programs changed:
oerep\r-invprtoe.w - suppressed dept printing on new Pacific invoice
oerep\r-invPrtProc.i - changed run program for new Pacific invoice (replacing old invPacif form)
oe/rep/invPacific.p - revision of xprint invoice form
- new logic at lines 391 to 409 to create summary temp table for invoice lines with matching (bol#, item#, po#)
- modified calculation/display logic to use new temp table instead of db table for inv-line
-does NOT change records in DB


brad-advantzware (changes requested):
Tough to code review the custom invoice but I will say that all invoices have the pre- and post-posted invoices.  The ar/rep/ folder needs a copy of invPacific.p that uses ar-inv...

PR #5015, opened 2019-08-02
Ticket - 51948 Advantzware TEST

sewa67:
Files - est/copypack.p
dEstPacking.w
est/CopyEnhancPack.i New File

### DIFF
--- a/Legacy/create.trg/estPacking.p
+++ b/Legacy/create.trg/estPacking.p
@@ -3,3 +3,6 @@
 TRIGGER PROCEDURE FOR CREATE OF {&TABLENAME}.
 
 {methods/triggers/create.i}
+
+ ASSIGN 
+    {&TABLENAME}.estPackingID      = NEXT-VALUE(estPackingId_seq) .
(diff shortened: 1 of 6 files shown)


brad-advantzware (changes requested):
dEstPacking.w - Change the list of supported material types to a global variable (constant) with INITIAL value of "C,5,6,M,D"
Change CopyEnhancPack.i to estCopyEnhancedPack.p and pass parameters (rowids)
Move the "bf-estPacking.estPackingID = NEXT-VALUE(estPackingID_seq) into the CREATE trigger of estPacking.p (find all instances of this and remove, once the create trigger is set.)

brad-advantzware (changes requested):
Sewa - The comment block at the top of CopyEnhancedPack.p was copied and pasted and is inaccurate.  You should have the comment block that gets auto created from PDSOE.  For a sample of that, see any of the new Procs.p programs (CostProcs.p).  Commenting in code is an area that we need to improve on.

PR #5026, opened 2019-08-06
Ticket -52817 Wrong item Break - Wrong Die Size - ink order

sewa67:
File - cerep/jobruff.p,jcrep/tickrrpt.i


brad-advantzware (commented):
This includes minor changes for McLean job card.  SHould not be an issue.

PR #5097, opened 2019-08-20
53222 Schedule Board add table sbJob

ron-advantzware:


brad-advantzware (changes requested):
"-" symbols in field names.  We want to avoid these due to complications with SQL based queries

PR #5104, opened 2019-08-20
Ticket - 53003 IF1- Unable to Export

sewa67:
File- fg/rd-fgexpN.w,jcrep/r-jobexl.w,oe/rd-invexp.w,oeinq/bol-expi.w,oeinq/rel-expi.w,rm/dRMExport.w,arinq/rd-invexp.w,
fg/vend-exp .w,oerep/rd-order.w,po/rd-poexp.w,fg/acc-exp .w,fg/carr-exp .w,fg/carr-mtx-exp.w,fg/fgnq-exp.w,fg/sman-exp.w,
arinq/csnq-exp.w,addon/browsers/rd-mtrexp.w,addon/browsers/rd-mh1exp.w,sys/ref/SelColCorrect.i

### DIFF
--- a/Legacy/addon/browsers/rd-mh1exp.w
+++ b/Legacy/addon/browsers/rd-mh1exp.w
@@ -940,6 +940,7 @@ PROCEDURE DisplaySelectionList2 :
       ldummy = sl_avail:DELETE(sl_selected:ENTRY(iCount)).
   END.
 
+  {sys/ref/SelColCorrect.i}
 END PROCEDURE.
 
 /* _UIB-CODE-BLOCK-END */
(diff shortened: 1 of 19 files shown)


brad-advantzware (changes requested):
Sewa Singh This appears to be all export programs.  What about handling this in selectable column reports?
Also, how can a user test this? Need to add steps to verify.  It may require changing data prior to using an export so include the use of Data Digger if necessary.
Additionally, your message is not to spec.  The message requested was:
"New Fields or columns have changed - Must delete and re-select columns desired"
Your inclusion of the field name(s) is good and necessary so please change the message to:
"Columns you have previously selected for inclusion have changed.  Please remove the following columns from your selection list and re-select the new columns:"  SKIP  [list of columns]

PR #5145, opened 2019-08-27
Ticket- 51950 On our Load Tags, we changed all the different shipping addresses to "Broker"

sewa67:
File - oerep/LoadtagProcs.p, oe/rep/lodxprnt9.i


brad-advantzware (changes requested):
"Customer Name" not changed to "Name"
Use of Global Variables is unacceptable.  There appears to be a temp-table that you use to populate data on these xprint loadtags.  Why would we not add fields to the temp-table and add the "broker" logic to the core program that populates the temp-table?  Don't overwrite existing fields, just use new fields and have the conditional logic to use those fields, in the lodxprint9.i program.

PR #5153, opened 2019-08-28
52909 Loadtag Maintenance Bin

brad-advantzware:


wade-advantzware (approved):
Looks correct based on the problem that was identified in assigning values to a loadtag from a 0 bin

PR #5165, opened 2019-08-28
52368 Price Matrix Logic

tyndmar:
Files changed: oe/PriceProcs.p, oe/PriceProcsLineBuilder.i
NK1 added - OEUseMatrixForNonStock - logical value = TRUE will use PriceMatrix for non-stock/custom items (defined in IF1)
In both changed files, called nk1look to retrieve switch value and process accordingly


brad-advantzware (commented):
As discussed, this will be approved since there will be a longer term refactor happening immediately after this.
Variable Definitions inside a block would normally be failed but I understand that Mark is compartmentalizing these changes to make it simpler to clean it up later.
Additionally, having one function/proc for "Settings"/Properties would be done in the "Main"/"Constructor" for the class/Persistent-proc.  This would include NK1 settings included within the procedure.  This Property could be simulated as a Public Function if there is value in knowing that setting externally.

PR #5168, opened 2019-08-29
Ticket - 53325 V-B-2 - Vendor not auto-populating as user types

sewa67:
Files - ap/w-selven.w


brad-advantzware (changes requested):
Please explain the hardcoded "3" in your code change.  On the surface it seems completely unacceptable (coded to pass the test case) but I'm hoping you have a good explanation.

PR #5170, opened 2019-08-29
Ticket - 53443 Allow UOM editing on O-B-1

sewa67:
Files -oe/d-invitm.w


brad-advantzware (changes requested):
Lookup, even though it was existing, should be changed to new Open Lookup look and feel.
May need to enhance that to allow for a pre-set list to be passed to it, if it does not already.
Validation should use Validate.p

PR #5171, opened 2019-08-29
Ticket - 51926 O-U-7 Loadtag Print Selection Screen is changing the Job# Suffix when estimate# is on multiple lines

sewa67:
Please review the change and merge into development branch
File - oerep/r-loadtg.w


brad-advantzware (commented):
Approving based on assumption that the job-no2 field is initialized properly and that this change corrects the overwriting of that correct initialization from the oe-ordl that is found.

PR #5205, opened 2019-09-03
Ticket - 37618 Error when user has reports in a directory and the directory is not set up

sewa67:
Files - cerep/r-mach.w
oe/rep/bolxprnt10.p
smartobj/lstframe.w
sys/Filesysprocs.p


brad-advantzware (changes requested):
You clearly copied and pasted code exhibited by the existence of comments that appear to relate to NotesProcs.p. 
You have duplicated code *within* the FileSysProcs.p (CreateFolder should *call* CheckFolder first rather than copying the same 2 lines in both places).
You have messages pertaining specifically to "BusinessFormLogo" inside a common function library.  This won't work.
Program should be in system\ not sys\.

I ask that you *think* about what procedures and functions other programs will need as it pertains for file system access. Think about inputs and outputs.  Any other file system routines that would be helpful to write once and then call as a procedure?

PR #5212, opened 2019-09-05
51875 LockMonitor enhancements

tyndmar:
Programs added:
util/wTxnLkMon.w
util/wTxnLkMon.wrx
see ticket for details


brad-advantzware (changes requested):
Change file name per Jay.
Create separate ticket to deprecate the programs this replaces.

PR #5213, opened 2019-09-05
Feature/53409 utility to recalculate jobs

wade-advantzware:


brad-advantzware (changes requested):
Rather than 2 loops through job-hdr, can we not assign the "after" costs immediately after jc-calc.p is called?
Additionally, I am ok with use of temp-table since output should be quick, but you should use the OutputProcs.p to generate the .csv.  Also need to confirm that c:\tmp folder is available.  Best to expose the choice for the log file location as part of the utility.

PR #5219, opened 2019-09-06
Ticket - 53636 A-F-1 Sales Rep Changes is not Updating All Files

sewa67:
Files - oe/d-invitm.w


brad-advantzware (changes requested):
Sewa - Validations should be using local variables with parameters that return valid yes/no and not rely on the ERROR-STATUS.
There is a merge conflict on this that looks resolvable, but it is not clear since global lMessageError and lError are both used.  Since you have to resolve the merge conflict, I also suggest pulling the latest devel and fixing the local validation procedures to pass back a parameter.

PR #5236, opened 2019-09-11
53507 ER4 Report

tyndmar:
Program changed:
cerep/r-prepN.w
removed "procat" (RM Category) from selectable fields


brad-advantzware (changes requested):
Spec changed to convert prep.procat to item.procat.

PR #5244, opened 2019-09-12
Ticket - 51949 Special payment terms

sewa67:
Files ar/invduedt,ar/v-arinv.w     - AU1 -
viewers/terms.w,methods/viewers/enable/terms.i   - AF4
util/w-arhead.w       - Util - Restore Invoice Purged via N-F-13
util/ar-inv-repair.p  -  Link not found
oe/rep/invandrop.p    - Nk1 - INVPrint unposted format - Androp
oe/rep/expgoodman.p   - Use in Aoa - Module OE - Menu Id -  OB4
oe/invhpost.i         - OB4 and Aoa - Module OE - Menu Id -  OB4
oe/rep/invconsb.p     - Nk1 - INVPrint unposted format - ConsBox
Util/dev/import.p     - Link not found
system/Commonprocs.p  - New File
ap/b-apload.w,       - VU1 
ap/v-apinv.w            - VU2


brad-advantzware (changes requested):
You did a pretty good job on this @sewa67 .  I think what I would prefer is that we keep the system\CommonProcs.p for the day of the month function but create a new system\CreditProcs.p that will encapsulate the entire terms table.  This can include a "GetInvoiceDate" function but it should also include at "GetTerms" procedure that returns the key terms fields ( discount days, discount percent, *and* due date - which calls GetInvoiceDate internally) given ipcCompany and ipcTermsCode.  Eventually, this can be enhanced to include all business logic related to credit checks and other credit hold functions.  These first public procedures would be the start.  Hold off on changes until we discuss the instantiation of these persistent procedures at standup.  These changes will be a lot simpler if we can just call the DYNAMIC-FUNCTION and should also allow us to remove programs like invduedt.i from the code base.

brad-advantzware (changes requested):
Sewa - The procedure should be run persistent, outside of the sub procedure or function.  You don't want to create and destroy the procedure every time you need to run it.  It should add once on load and remove once when done.

PR #5302, opened 2019-09-20
Ticket - 54058 A-F-1 Sales Rep Changes is not Updating misc Items on invoices (Unposted)

sewa67:
File - oe/d-invitm.w


brad-advantzware (changes requested):
Holding.  I don't think this was desired spec.  Spec was not provided.

PR #5518, opened 2019-10-23
54392 VendItemCost Table - Data Conversion on Upgrade

tyndmar:
File changed: asiUpdateEnv.w - modified VendorCostConversion processes to conform with new inline procedures


brad-advantzware (changes requested):
RUN pRecalculateFromAndToForLevels(bf-vendItemCost.vendItemCostID).  is still called from within here.  Is this a reference to another private proc?  I don't see it. 
Question:  Does this have to be a monolith because you don't have propath when running?  I think this is going to get bigger and bigger and having a .p that you can call will be far easier to manage.
My concern is that there are temp-tables and global variables that only apply to one conversion.  Ideally, this is all encapsulated in one .p

PR #5537, opened 2019-10-28
Ticket: #55447 Invoice Posting - Use of RECID may be problematic

vishnuvellankijkt:
Files:
oe\r-inve&pN.w


brad-advantzware (changes requested):
RECID->ROWID  
second instance of assignment is inv-line, should be inv-misc

PR #5584, opened 2019-11-01
55800 Error entering order: ** OE Release Master already exists with …

tyndmar:
…r-no 37521. (132)

File changed: oe/oe-rel.a
Added loop if seqence returns a previously used r-no
This (1) acquires a valid r-no for the release and (2) updates the sequence so next number will be properly incremented


brad-advantzware (changes requested):
move uniqueness validation inside the procedure

PR #5602, opened 2019-11-05
5730 Military Time vs. AM/PM

ron-advantzware:


brad-advantzware (changes requested):
Ron Stark I could not get a hold of you to review this with me.  Ideally, I would like to not have to contact you to walk through your code changes, so I have to fail the code review.
As per process, each program (or groups of programs) should have a quick summary of the changes within each.  Using more comments in code can also help me and future developers.
In general, I understand what your doing with the common function and that is great.  But this shoudn't be in the session, it should be in Common.  Adding Mark Tyndall to see if he would prefer to wait until he organizes this before requiring that change.
I believe I understand all cases where you are running the time display through your new function.  That seems good.  This should also allow us to more easily handle time zone switching at some point, too, if all time is displayed through this function (I hope).
However, I don't understand what the changes to the the include files impacts.  Need an explanation there.
Also I don't understand why are there blocks of code deleted in many of the programs.  I can't tell if this is conscious decision or if purged by an AppBuilder.

PR #5605, opened 2019-11-05
Ticket - 54620 AU1-Blank Invoices

sewa67:
Files - ar/b-arinvl.w
ar/r-arve&p.w
ar/w-arinv.w
oe/r-inve&pN

### DIFF
--- a/Source/AOA/BL/r-inve&pb.p
+++ b/Source/AOA/BL/r-inve&pb.p
@@ -2376,6 +2376,8 @@ PROCEDURE run-report :
         AND inv-head.cust-no  LE cEndCustNo
         AND inv-head.inv-date LE dtEndInvoiceDate
         AND inv-head.stat     NE "H"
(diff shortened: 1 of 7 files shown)


brad-advantzware (changes requested):
Sewa Singh For each file that you change, you must include a note about what you changed and why.
For instance, why did you remove the "clear-ar" procedure call?
I see you moved the locaiton of the ar-invk.i - is this to calculate the tax after delete?
Why did you do have an additional "add-link"
*PLEASE* save me the time of having to open the code and investigate myself and prevent me from failing your code reviews.

Additional changes required:
Change message from: "The last line item is deleted, are you sure to delete Invoice Header ?" to "You have deleted the last line item for this invoice.  Do you want to delete the invoice header as well?"
There are other posting programs in AOA/BL and AOA/dynBL that should be changed to be consistent.

PR #5616, opened 2019-11-07
Ticket - 56082 The routing or ABA number is an integer in the db, butneeds to support a preceeding zero value

sewa67:
Files - viewers/bank.w - corrected the routing format
util/ImportVend.p  - Added New field and match with export
Util/ImportCust.w - Added new field and match with export
fg/cust-exp.w     - Export data in xls format
fg/vend-exp.w    - Export data in xls format
custom/d-bankinfo.w  corrected the routing format
Resources/template/ExportData.xls - template with blank string format .


brad-advantzware (changes requested):
Recommit with only the changes to viewers and importers.

PR #5641, opened 2019-11-11
Ticket - 55984 Cleanup of zMessages

sewa67:
Files - browsers/probe.w
ce/probe.w
custom/comp_loc.w
pc/r-wipedt.p
pc/r-wippst.p
po/d-poordl.w
system/MessageProcs.p
System/v-message.w
ce/v-est3.w
cec/v-est3.w
custom/d-prompt.w


brad-advantzware (changes requested):
Closer but not where it needs to be if this is to be a common function in out system.

PR #5676, opened 2019-11-14
Ticket - 56724 Whse/Bin Locations - user can save non-existent Bin on Location header.

sewa67:
Files - viewers/loc.w


brad-advantzware (changes requested):
See InventoryProcs.p  Review existing Validation programs and use those.
Create new "AddNewLocation" procedure there as well.  
Remember - avoid business logic in window/UI programs

PR #5680, opened 2019-11-14
Ticket: #54690 API Inbound - CreateTagForPO - FG

vishnuvellankijkt:
Files:
api\inbound\handlers\CreateTagForPORequestHandler.p.p
api\inbound\CreateTagForPO.p


brad-advantzware (changes requested):
"KELLOGS" should not be hard-coded.  If neither the po-ordl.cust-no or oe-ord.cust-no, then cust-no is "".
Need to test availability of oe-ord before using oe-ord.cust-no.
"cust-par" table is "cust-part".  this shoudl still compile but we shouldn't shorten full DB table names.

PR #5690, opened 2019-11-18
Ticket - 55207 Custom Import - RM Items and Beginning Balances

sewa67:
Ticket - cec/w-item.w
rm/dRMExport.w
util/Validate.p
util/ttImport.i
windows/item.w
util/ImportItem.p
util/ImportItemFold.p,
util/dev/impItem.p
util/dev/impItemFold.p


brad-advantzware (changes requested):
Sewa - I don't want 2 copies of basically the same import program.  Add an Industry field to the import so that  you have one Raw Material Item Importer.
This should be true of the Export as well.  Are there any major differences between the two?

PR #5712, opened 2019-11-21
Ticket - 49523 O-U-1 Steel "Die Rule" costs on Order Misc Tab changes all Forms after Vendor PO is created

sewa67:
Files - oe/oelibupd.p
oe/ordlmisc.p
oe/v-ord.w
po/updordpo.p


brad-advantzware (changes requested):
Sewa - The blank-no (b-no) is not being assigned which was requested in the spec.

PR #5726, opened 2019-11-25
50569  revised commit

Umax62:


brad-advantzware (changes requested):
@Umax62 This is supposed to have been done on a NEW feature, which is why we created additional tickets for this.  This should've been done under 56901.
Cancel this PR, create a new feature tied to ticket 56901, and resubmit this pull request for that feature branch.
You also are not recording time entry on this new ticket which you must begin doing immediately.

PR #5779, opened 2019-12-05
Ticket - 54026 GPI- Staunton Pick Ticket

sewa67:
Files - oe/rep/relcardx.i
oe/rep/relcardx2.i


brad-advantzware (changes requested):
Spec change from customer

PR #5806, opened 2019-12-11
Ticket - 47394 Unassembled sets not showing cost on View Item tab

sewa67:
Files - fg/d-rcptd.w
fg/fg-post.i


brad-advantzware (changes requested):
Sewa -  You have no explanation as to the change to fg-post.i nor why creating a buffer will fix the error.

PR #5850, opened 2019-12-17
Ticket - 44055 Address sold to/ship to

sewa67:
Files - ar/rep/invlovepac.i
ar/rep/invlovepac.p
oe/rep/bolxprn10can.i
oe/rep/bolxprnt10can.p
oe/rep/invlovepac.i
oe/rep/invlovepac.p


jay-advantzware (commented):
Only printing partial address, but cannot print full address due to size of full address.  Seems ok

PR #5859, opened 2019-12-18
Ticket - 24677 NK1-BOLPartial - determines the BOL P/C value based upon Order or Release

sewa67:
Files - oe/oe-bolpc.i  - logic of  BOL P/C value based upon Order or Release
sys/inc/create_nk1.p
sysref/sys-ctrl.i


brad-advantzware (changes requested):
See comment in ticket.  Need to review oe/oe-bolpc.p since it runs after this logic.  Code change is likely better done within the .p

PR #5880, opened 2019-12-20
Ticket - 58754 IF1-False PO Status Indication

sewa67:
Files - viewers/itemfg.w


ron-advantzware (changes requested):
The iCount field seems to be a display only value.  It's Enabled property should be set to NO and the DISABLE iCount line in procedure enable-itemfg-field should be removed.  Going to NY by way of San Fran.  Jumping through hoops to disable this field in various places when it simply should not be enabled to start with.

PR #5907, opened 2019-12-24
Ticket - 59338 Revision to "Delta2" BOL Format

sewa67:
Files - oe/rep/boldelta.i
oe/rep/boldelta.p


brad-advantzware (changes requested):
Sewa - is the "WHEN AVAIL" oe-ordl necessary when displaying cJobNo?

PR #5959, opened 2020-01-03
57218 Detailed Transactions user folder missing

ron-advantzware:


jay-advantzware (commented):
This should not be using new file procs?

brad-advantzware (changes requested):
Why is the "users" directory hard-coded to be used here?  Is the intent to get a temp folder?  If so, GetTempDirectory from FileSysProcs.p should be used.

PR #5963, opened 2020-01-05
58928 IF4 Import

tyndmar:
New/changed programs:
util/ttImport.i - added support for new program ImportLoc.p
util/importLoc.p - new program to load loc and location tables from csv template


brad-advantzware (changes requested):
Is this a duplicate to ImportWarehouse.p?  This may be a huge waste of time if so.

PR #5971, opened 2020-01-06
Ticket: #51737 Vendor EDI Kickoff

vishnuvellankijkt:
Files:
api\node\InboundAPIRoutesJSON.js
write.trg\ap-inv.p
api\inbound\CreateVendorInvoice.p
api\inbound\handlers\CreateVendorInvoiceRequestHandler.p


jay-advantzware (commented):
Approved - Just needs to be tested - already loaded the API data, so need to test.

PR #5976, opened 2020-01-07
Ticket - 60179 RM Importmaster Fields and usage

sewa67:
Files - util/ImportItem.p


jay-advantzware (commented):
Think these fields need to be validated and cannot tell if this does that.
Brad to review.

brad-advantzware (changes requested):
Anything required should also be validated if it is based on a set list.
Procat and mat-type need to be moved from the *optional* validation and be part of the required field validation.  Must add validation for cost-type in the required field section.

jay-advantzware (approved):
Approved

PR #5978, opened 2020-01-07
59593 Invoicing MISC Items from OU1 not working

tyndmar:
Program changed: oe/b-ordm.w
Fixed issue with adding more values to Billing Instructions than permitted by field extent limit
(was trying to write 5, only 4 allowed)


jay-advantzware (approved):
Approved

PR #5989, opened 2020-01-08
Ticket: #60270 MF2 Purchased UOM Problem should NEVER be unable to Update regardless of ON HAND

vishnuvellankijkt:
File:
cec\v-item2.w


jay-advantzware (approved):
Approved

jay-advantzware (commented):
Approved

PR #5991, opened 2020-01-08
60318 IF4 Upload

tyndmar:
Files changed/added:
util/ImportWarehouse.p - correct designation of active flags to ensure correct value set
util/impWhseScript.p - throwaway program to be used to load data on Premier version 16.12


jay-advantzware (commented):
Not sure what all this is other than the first program?  Need Brad to review

PR #5996, opened 2020-01-08
60598 Each check printed is advancing down a line

tyndmar:
Files changed:
ap/ap-delta.p - removed last line feed to prevent form creep


jay-advantzware (commented):
Approved

PR #6006, opened 2020-01-09
Ticket - 60604 Incorrect Sq Ft Totals on O-Z-1 Report

sewa67:
Files - oerep/r-sqrshipN.w


jay-advantzware (commented):
looks ok - But i would like to start pushing more on black box methods for things - like building blocks for calcs and such - Like to get MSF (Send length, width, size uom, quantity, rounding) etc.  These may be available in tbox to leverage.  Not today, but in all new stuff for sure.

PR #6036, opened 2020-01-11
Ticket - 60473 Load Tag Format - Logo, NoLogo

sewa67:
Files - oerep/LoadtagProcs.p
oe/rep/lodxprntNoLogo.i
oe/rep/lodxprntLogo.i


jay-advantzware (commented):
This seems like we could create a customer driven load tag format and if it is done like this, AP checks would be similar and likely easier since fonts are limited to 2.

PR #6042, opened 2020-01-13
Ticket: #60939 Amazon Order Custom cXML Integration (Enhancements - Use Site Id / Shipto Location to get FG Item #)

vishnuvellankijkt:
Files:
cXML\gencXMLOrder.p
cXML\getCustDetails.p


jay-advantzware (commented):
Procedure to getPart - Could we make 1 call to get this or do we need to try multiple times? They need the specific ItemFG based on the SIteID as there will be duplicates of just customer part number, so concerned that maybe this procedure needs a GetPart(Method of getting part) or something?

PR #6060, opened 2020-01-17
58546 Enhance Rec_Key Assign

ron-advantzware:


brad-advantzware (changes requested):
Add Stack Trace to this Super-not-running case

PR #6062, opened 2020-01-17
Ticket - 60864 Various changes for UI / Operation of Vend Item Cost From Estimating

sewa67:
Files - viewers/vendcostmtx.w


jay-advantzware (commented):
Not sure if this will be applied to normal Add or just copy.  We only want this to be on Copy and want these defaults on Add.

PR #6082, opened 2020-01-23
Ticket: #61519 API Inbound - Failure validation and responses

vishnuvellankijkt:
FIles:
cXML\gencXMLOrder.p
cXML\getCustDetails.p


jay-advantzware (commented):
Would like better description on line 141 - XML is not in valid format is not the case - It is that the payloadID = Blank or Order Date is not defined, or OrderID is blank, or Ship to Id is blank.  We are trying to give very clear message so they can find the issue and correct it without contacting us.

jay-advantzware (commented):
Logic for part number - First check at line 188 is good and message is accurate as the data is missing from the file submitted.

Line 533 - I thought we had agreed NOT to add details on the line numbers to this check in because the output may not be supported?

PR #6140, opened 2020-02-03
Ticket - 62063 JU1 Mass Delete - Add Select All and Deselect All option

sewa67:
Files - jc/JobItemPop.w
jc/b-jobmat.w
jc/b-jobmch.w


brad-advantzware (changes requested):
Sewa -  
Your temp-table definitiion must be in a .i include file.  You should have the NEW SHARED as {1} argument to also allow this to NOT be a shared temp-table (not a required change right now)

PR #6152, opened 2020-02-04
Ticket - 53252 Label Matrix - Control Reanalyzing with a forced reanalyze (add /Z=1 command)

sewa67:
Files - custom/Imprint.p
sys/inc/create_nk1.p


brad-advantzware (changes requested):
This looks OK but the /Z= command should be *appeneded* to the end of the command string.  It looks like you are replacing the /L= command.

PR #6164, opened 2020-02-05
Ticket #62282 Event View - Response Result: insert rather than append on Resubmit

mithunporandlajkt:
Files:
api/APIInboundEventsViewer.w
api/CreateAPIInboundEvent.p


jay-advantzware (approved):
Approved

PR #6185, opened 2020-02-07
Ticket - 61038 N-M Update Order Utility Runs Infinite Loop / Defect

sewa67:
Files - oe/ordlsqty.p
util/fxoeship.p


brad-advantzware (changes requested):
Sewa - You change to require match on ar-invl.ord-line worries me.  Where is that value assigned?  Very concerned that this will result in 0 invoiced quantity. Did you run this through a test on DEVEL?

PR #6224, opened 2020-02-13
57232 - AutoISSU - JobClose - Problem with UOM Cost

brad-advantzware:


jay-advantzware (changes requested):
Whenever doing division, like / qty, need an error statement or capture the divide by zero possibility.  Is that handled here?

PR #6373, opened 2020-03-16
Ticket - 64506 IF1- Adjustment Transactions Not Calculating Cost Correctly.

sewa67:
Files - fg/b-fgadj.w


brad-advantzware (changes requested):
See ticket comment for questions

PR #6380, opened 2020-03-17
Ticket - 61572 Request - Add PO# to IC1 Report

sewa67:
Files - fgrep/r-cycntN.w


jay-advantzware (commented):
This will not work for PO where the same item is on the po more than once, but not sure if we can get it correctly elsewhere?

PR #6421, opened 2020-03-24
Ticket: #64838 PF3- Vendor Cost Matrix browse - Needs to limit initial load

vishnuvellankijkt:
Files:
browsers\vendcostmtx.w


jay-advantzware (commented):
Do not see the default integer value set to 25?

PR #6466, opened 2020-03-31
64563 Request - Inactive Flag on Machine File

rahulrawatjkt:
Files:
est/d-estop.w
schedule/objects/resources.w
touch/allmachs.w
touch/machines.w
windows/l-mach.w

### DIFF
--- a/Source/est/d-estop.w
+++ b/Source/est/d-estop.w
@@ -1938,8 +1938,7 @@ PROCEDURE valid-mach :
 
         IF mach.obsolete THEN 
         DO:
-            MESSAGE "Machine is Inactive, please choose a different machine"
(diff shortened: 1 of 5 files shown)


ron-advantzware (changes requested):
When using a logical field in a WHERE phrase, it is a best practice as recommended by Progress and Dan Forman to not simply use WHERE <LogicalField> without explicitly adding the comparison component.  Please can WHERE <LogicalField> to WHERE <LogicalField> EQ YES and WHERE NOT <LogicalField> to WHERE <LogicalField> EQ NO.  Without the comparison component, Progress does not utilize the <LogicalField> if part of an INDEX and you may inadvertently cause Progress to perform sequential search.

ron-advantzware (approved):

PR #6548, opened 2020-04-15
62968 EDI - Tri-Lakes (Welch)

tyndmar:
Programs changed:
po/po-hrexp.p
- almost a complete rewrite
- driving issue was lack of reference to hrms reftable (removed as unneeded for valley)
- put this back, plus minor changes from conversations with EFI, Welch, TriLakes
sys/inc/hrms.i
- added reference to NK1 "HRMS" integer value
- must be set to "1" to use reftable; this to prevent unplanned changes to Valley transmissions

### DIFF
--- a/Source/po/po-gpexp.p
+++ b/Source/po/po-gpexp.p
@@ -4,95 +4,100 @@
 /*                                                                            */
 /* -------------------------------------------------------------------------- */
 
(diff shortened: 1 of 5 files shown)


jay-advantzware (commented):
This has hard coded customer id and does not seem to be API driven?

PR #6557, opened 2020-04-17
Ticket - 66667 Entry of order - Using Estimate with a Source Estimate ID

sewa67:
 Files - oe/d-oeitem.w


jay-advantzware (commented):
Does a normal estimate match on FG item # or can the user enter the FG item number on the order line after linking to the estimate?

PR #6702, opened 2020-05-14
Ticket - 66302 Board Ordering - Via an import/export of purchase orders

sewa67:
Files - util/ImportPo.p


jay-advantzware (commented):
Sewa - For new items we will use the standard of <AUTO> to create a new PO with a new number and <AUTO>##### where the user can enter a group number to add multiple lines to the same PO number.  This is the same as the quote import logic.

PR #6707, opened 2020-05-15
68657 Replace this system message with a NK1 such that this message can be suppressed

rahulrawatjkt:
Files:
methods/prgsecur.i
sys/inc/create_nk1.p


jay-advantzware (commented):
Spelled in correctly - INPUT "To show or suppress the restrcition message"
'restriction'

PR #6730, opened 2020-05-18
68513 IU5- Slow posting time

tyndmar:
File changed:
fg/fgPostBatch.p
- replaced REPEAT loop with DO WHILE with embedded 1 second pause
- added user messaging to advise of issue encountered


jay-advantzware (commented):
NO - We do not want to lock FGItem at all nor send a message.  Do a no wait, no error and if item fg cannot be updated, skip the update and proceed

jay-advantzware (commented):
Should code at line 666 be removed as duplicate code?

jay-advantzware (commented):
Can we release before sending email (If setup)?

jay-advantzware (commented):
Looks good.

PR #6884, opened 2020-06-12
69173 Remove Rec_key table

ron-advantzware:


brad-advantzware (commented):
I would prefer to delete rather than comment out just so it doesn't come up on grep but not worth going back to edit probably.

PR #6916, opened 2020-06-17
70854 Schedule Board - add Job Quantity, Est Score and correct PO values

ron-advantzware:


brad-advantzware (changes requested):
Ron - See inline questions/comments

PR #6996, opened 2020-07-01
64851 - Create utility to Purge orphan records

tyndmar:
File Changed: util/purgeProcs.p
Improved reporting to better identify records to be purged
see sample in ticket


jay-advantzware (commented):
Change for tax is for #70670 - Looks good.

PR #7007, opened 2020-07-02
73458 Error Message when trying to post FG Physical Analysis & Posting

tyndmar:
Change loop from EXCLUSIVE to NO-LOCK and added single-record lock to avoid Lock Table overflow
Change FOR EACH to remove internal LOOKUP statement on left of operator and moved to loop interior to avoid full table scan
Removed EXCLUSIVE in second loop (line 1589) since no DB activity inside loop


jay-advantzware (commented):
Looks ok - but do not see a new lock to replace exclusive, so assume that is handled.  
1) Should line 1577 - should that be the buffer?
2) Should the row id be the buffer?

PR #7022, opened 2020-07-07
#71500 UpdCust1 - Enhance and improve

rahulrawatjkt:
Files:
ar/r-cshe&p.w
ar/updcust1.p


jay-advantzware (commented):
Why do the exclusive lock on cust if we are not sure we need t update?

PR #7042, opened 2020-07-11
70223 DAOA - Output to csv to not require Task Scheduler or Jasper st…

ron-advantzware:
…arter to run


jay-advantzware (commented):
Can we compile and view results in override before checked into Source code?  The new exports are not usable without export to CSV and do not want to require Jasper starter installation on all workstations.

PR #7065, opened 2020-07-15
Ticket - 67807 N-K-1 SSBolEmail - Logical value = No Sharp Shooter BOL Creation Adding Items to Order That Are not on Release

sewa67:
Files - addon/bol/b-relbol.w
adddon/bol/b-updrel.w


jay-advantzware (commented):
Looks good - The blank password removal is a change and if not needed to be removed, then prefer it stays as to keep existing functionality

PR #7128, opened 2020-07-28
91312 0-ar cXML Fixes

brad-advantzware:


jay-advantzware (commented):
invNoStart AS INTEGER FORMAT ">>>>>>>9" - 1 digit more than InvNoEnd - not needed but does not seem to be a problem.

PR #7250, opened 2020-08-19
91989 DSG rejection notifications - SUFFIX

tyndmar:
programs changed:
oe/rep/invpremx.p - added display of invoice suffix if assigned in OB5
oerep/cXMLCInv.w - added field for invoice suffix if retransmitting


jay-advantzware (commented):
1) Why check length = 2?  Anything they enter should be considered data
2) Remove message

PR #7264, opened 2020-08-21
#91941 Finalization and adjustments to Vertex

rahulrawatjkt:
Files:
cXML/cXMLInvoice.p
oe/rep/invpremx.p


jay-advantzware (commented):
Approved

jay-advantzware (commented):
Need to review invoice with tax on freight and verify totals are accurate - this does not seem totally accurate, but cannot tell from code.

PR #7414, opened 2020-09-22
Ticket - 91638 zMessage to alert user if prep item was disposed

sewa67:
FIles - est/BuildDefaultPrep.p
jc/jc-calc.p
system/PrepProcs.p
oe/ordlmisc.p


jay-advantzware (commented):
Will this only display when the item is initially added - Would not want this to appear every time an order is updated, only on initial add.

PR #7525, opened 2020-10-20
Ticket - 93727 IU2/OS1- Back Order Transfer

sewa67:
Files - oe/bol-pre-post.p


jay-advantzware (commented):
Should this be <> 0 to support returns?

PR #7539, opened 2020-10-22
Ticket - 93176 Add Expected Revenue Field to the Bil of Lading

sewa67:
Files - oe/bolhtots.p
oeinq/bol-expi.w


jay-advantzware (commented):
On a bolh - for each boll - why not just go to get the price from oe-ordl directly to get price?
Where is the calculation to get the total price when we create an invoice?

PR #7893, opened 2021-01-07
95990 Fix compile errors 01/06/21

tyndmar:
various issues arising from NewUI includes


jay-advantzware (commented):
Do not see why this was added?

PR #7954, opened 2021-01-18
96181 Deposits Not Showing in VT1 and VT2

tyndmar:
File changed: asiUpdateENV.w
Fixed issues where (1) Correct DataFix procedure did not run and (2) if it HAD run, would throw record not found errors due to LOCK/NO-LOCK conflicts


jay-advantzware (commented):
Approved

PR #7979, opened 2021-01-20
#93360 Enhance Order entry browse to display current records first, add limit to query to prevent delayed search

rahulrawatjkt:
Files:
browsers/BrowserProcs.p
oe/b-ordinq.p
sys/inc/create_nk1.p
system/d-QueryLimitAlert.w
system/session.p


jay-advantzware (approved):
Too much to review - passed testing in feature branch

PR #8021, opened 2021-01-22
#62604 GL Account – Do not allow inactive account number entry anywhe…

seemaranijkt:
…re in system_2

Files:
custom/ar-ctrl.w
glrep/r-distrb.w
oe/d-inmisc.w
oe/d-ordm.w
viewers/arclass.w
viewers/costtype.w
viewers/fgcat.w


jay-advantzware (approved):
Needs to be tested - but looks better

PR #8038, opened 2021-01-26
67465 Department Codes (also 96500 Deploy prgrms images)

tyndmar:
Program changed: asiUpdateEnv.w
added logic on every upgrade to set department isRequired flags
added procedure ipSetDepartmentRequired
used list: "PR,GL,QS,WN,WS,FB,FS,RC,RS,CR,PR,CT"
change assignment of menuimage (without extents) to menuimage[1]


jay-advantzware (commented):
Do not make them active if they have made them inactive already - bDept.isInactive = FALSE.

PR #8061, opened 2021-01-28
64851 Create utility to Purge orphan records - PO Files

tyndmar:
Files changed:
util/purgeOrphans.w - turned on options for Purchasing files
util/purgeProcs.p - added support for po-no checking


jay-advantzware (approved):
Approved

PR #8067, opened 2021-01-29
96626 Cannot add a vendor cost in IF1

sachin448:
File - fg/v-eitem.w


jay-advantzware (approved):
Removed unnecessary procedure

PR #8077, opened 2021-01-29
96433 SS FG Inquiry must have two Fields for Job and Suffix

rajeshadv:
File
1. sharpshooter/b-fgInqBins.w


jay-advantzware (approved):
Approved

PR #8087, opened 2021-02-01
94465 TR! - Beginning Employee Field Can't Be Blank, but Has No F1 Lookup

sachin448:
Files - touch/r-emptrn.w


jay-advantzware (approved):
Passed testing in branch.  Merged.

PR #8104, opened 2021-02-04
Ticket - 96586 J-U-1 FG Creation Mod

sewa67:
Files - jc/fgPrompt.w
jc/fgadd.i
jc/jc-calc.p
jc/v-job.w


jay-advantzware (commented):
Why not use buffer name here: define buffer itemfg for itemfg
Is the fgadd.i a logical thing to make into the procedure to add FG item?
What does FG file maintenance do differently?

PR #8260, opened 2021-03-01
97604 Expand Job# Field. When typing Job# W01038, the entire job number does not appear

sachin448:
File - Source/pc/b-prdd.w


jay-advantzware (approved):
approved

PR #8291, opened 2021-03-04
97189 AF1 Industry Field

goutamsharmajkt:
Created a new maintenance for the industry and added in the A-F-1 screen


jay-advantzware (commented):
Need triggers added for the new table?

PR #8300, opened 2021-03-05
97450 Quote export not working / PO on G/L

tyndmar:


jay-advantzware (commented):
Profit is / $0.00 is likely still going to fail - need to handle exceptions and make sure they do not cause an error if we use any format for data.
Ideally a procedure that we pass (pProfit(->>,>>9.99%) would return a valid value or the upper/lower limit.  For example -100000000000 would return -99,999.99% and same for high value.  If we have this global procedure, then anytime we get this error, we can add procedure to handle it.  Otherwise, we end up putting little logic checkers everywhere.

jay-advantzware (commented):
RUN get-itemfg-gl (INPUT cocode, INPUT SELF:screen-value, OUTPUT cActNum).
- There is an NK1 = Ap GL# that needs to be used here, so if blank, should not prefill it as it likely is blank in the source defined.

PR #8414, opened 2021-03-30
98252 MSF - Reset quantity

tyndmar:
Program changed: asiUpdateENV.w
added script to update zero gshQtyInSF from tot-lbs


jay-advantzware (approved):
Merged as need to test upon upgrade

PR #8769, opened 2021-06-14
65250 System changes COA posting account during Posting Process

sewa67:
Files - Source/ap/r-apve&p.i


brad-advantzware (approved):
REviewed with Sewa - this will prevent reassignment of the account number and just use the ap-invl account number determined inside VU1

PR #9222, opened 2021-10-07
103999_Refactor_newblank_One_program_across_industries

Pandianrajjkt:
refactor newblank to NewEstimateBlank and change the calling programs respectively


brad-advantzware (changes requested):
Varun - A couple changes - Is there a reason why you removed the internal procedures GetQtyP etc and moved those into one big procedure?  If you are going to remove code, you can just delete instead of commenting.   Commenting instead of deleting can be helpful as  documentation but only if the logic changes significantly (and additional clarification is added).  In a cleanup/reformatting, deleting is preferred since it doesn't really add value.  We can always look at github history.
Another change:
There seem to be 2 different blocks where the .est-type of 2 or 6 is conditional.  Could this be combined into 1 block?

PR #9234, opened 2021-10-12
102290_Plates_on_combo_estimate_dropping_off_from_prep

sakshisinghjkt:
Removed plate deletion code from write trigger. Added one new method into PrepProcs.p and moved logic to validate and delete Plates.
File: 
Source/write.trg/eb.p
Source/system/PrepProcs.p
Source/ce/b-estitm.w


ron-advantzware (changes requested):
You added local-destroy for Prep Procs, however, there are 3 others previously instantiated, please add them to the local-destroy

PR #9281, opened 2021-10-26
104474 System Configuration Setting

rajeshadv:
Files
1. system/Setting.cls
2. viewers/setting.w


ron-advantzware (changes requested):
No hard coded security levels should exist outside of session.p.  There are functions in session.p to test for Admin level users and Super Admin users which return a logical.  This was done this way so if security level values change in the future, the only place requiring a change would be session.p.  Please adjust the code to utilize DYNAMIC-FUNCTION("sfIsUserSuperAdmin").

PR #9290, opened 2021-10-28
61757 - Burt Rigid Jobcard - new JobCardF 20 format

rajeshadv:


ron-advantzware (changes requested):
Program cerep\jobburt.p - please review procedure pGetJobQty and note bff-oe-ordl at lines 746 and 752 does not reference a record LOCK status (thus implied SHARED-LOCK).  Please set to NO-LOCK and/or EXCLUSIVE-LOCK accordingly.  Also, as a general practice, we prefer the table LOCK status on the same line as the FIND, FOR EACH etc, and before the WHERE phrase.  Since you are making a change as outlined above, please review any other table access statements and reposition the LOCK as recommended above.  We do this, so when performing a grep on a specific table looking for EXCLUSIVE-LOCK when trying to trace TRANSACTION scope, it's so much easier to eliminate opening a program file when it is readily seen the table being accessed has a NO-LOCK phrase.  Thanks.

PR #9329, opened 2021-11-15
104996 SendJobAMS - Add Customer and Quantity information at Order level

rajeshadv:
File
1. api/SendJobAMS.p


ron-advantzware (changes requested):
There is no LOCK status phrase explicitly expressed on table job-hdr line 481.  Please add the lock status phrase.

PR #9340, opened 2021-11-17
103168 Job Allocation For Distribution MOD-ID: 10169

sewa67:
Please review this change and merge into development branch .
Files - Source/jc/jc-calc.p
Source/jc/v-job.w
Source/oe/d-oeitem.w
Source/oe/ordlup.p
Source/oe/v-ord.w


ron-advantzware (changes requested):
Appears a debug message left in oe/v-ord.w at line 2673, please remove.
Also, any reason using windows/l-loc.w in jc/v-job.w instead of using a dAOA lookup?  We should be striving to utilize any and all lookups using dAOA.

PR #9341, opened 2021-11-17
105041 NK1 to NK6 Conversion for Zoho

sachin448:
File - Source/CRM/crmContacts.i
Source/CRM/crmCustomers.i
Source/CRM/w-genRefreshToken.w
Source/CRM/ZohoProcs.p


ron-advantzware (changes requested):
In CRM/ZohoProcs.p, after instantiating oSetting at line 30 you need to add
oSetting:LoadByCategoryAndProgram("ZoHoCRM").
You should be adding this call everywhere you are referencing NK6 values.  This allows us to see what settingType and setting table values are potentially being used in each module.

Whenever you are replacing direct sys-ctrl access, we are no longer auto creating the NK6, like we used to do with NK1's records.  So, please remove oSetting:Update statement at line 101.  For the time being, when an NK1 to NK6 ticket is assigned to you, the settingType record will already exist and setup by me.  If it is not there when you get an NK1 to NK6 ticket, please inform be immediately before working the ticket.

In CRM/w-genRefreshToken.w, at line 55 add oSetting:LoadByCategoryAndProgram("ZoHoCRM").  Again, always attempt to place this type of call whenever you are adding these oSetting = NEW calls.  If you have any questions, etc.  Don't hesitate to ask for clarification from myself, or a better option might be to ask Rajesh.

PR #9364, opened 2021-11-23
105015 DSG BOLs Backlog

tyndmar:
Programs changed:
AOA/dynBL/r-bolpst.p - added logic to update oe-rel.spare-char-1 if BOL has been assigned to a different warehouse
added logic to run fifoLoopTags for the BOL warehouse instead of the original ship-from from the order
util/ClearDSGTags.w - utility to remove tags from BOLs when needed


ron-advantzware (changes requested):
The addition of oe-rel EXCLUSIVE-LOCK at line 386 turns the entire Procedure pAutoSelectTags into a TRANSACTION block.  There are two reiterative loops in this procedure which if a large client with large amounts of data could blow out the LOCK TABLE Stack.  Please wrap the two FIND CURRENT statements in a DO TRANSACTION block.

PR #9374, opened 2021-11-29
105050 Location shares the same address between companies

tyndmar:
Programs changed:
browsers/loc.w
custom/companyCopy.w
fg/loc-exp.w
(see next comment for details)


ron-advantzware (changes requested):
Please review several table where  no LOCK phrase is not explicitly entered.  There are several places where implicit SHARE-LOCK which upgrade to EXCLUSIVE-LOCK.  Based on our coding standards, the lock phrase should be explicit for clarity, especially for junior developers down the road.

PR #9411, opened 2021-12-06
105169 Estimate form change request

sachin448:
File - Source/cec/probe.f
Source/cec/probev1.p


ron-advantzware (changes requested):
CostHeader has an EXCLUSIVE-LOCK.  There doesn't seem to be any need to LOCK this record.  Please change the LOCK phrase accordingly.

PR #9421, opened 2021-12-08
105621 Create AMS Job Loaded Utility

sachin448:
File - Source/util/loadAMSJobs.w


ron-advantzware (changes requested):
Add logic to DELETE PROCEDURE for instantiated PERSISTENT procedures found at lines 47 & 50.  Whenever instantiating a PERSISTENT procedure(s), properly removing them is required to avoid performance and blowing memory out.

PR #9422, opened 2021-12-08
105623 Two AR reports differ now

sewa67:
Files - Source/glrep/r-curdetA.w
Source/glrep/r-curdetN.w


ron-advantzware (approved):
You could save some typing by combining the two ASSIGN lines when setting a variable based on a conditional at lines 874-877 in glrep/r-curdetA.w
dtPerStateDate = IF AVAILABLE THEN period.pst ELSE tran-date.
Same goes for lines 1336-1339 in glrep/r-curdetN.w

PR #9423, opened 2021-12-08
105632 Close Open POs

sewa67:
Files - Source/util/ClosePurchaseOrder.w
Source/util/ClosePurchaseOrderProc.p

### DIFF
--- a/Source/util/ClosePurchaseOrder.w
+++ b/Source/util/ClosePurchaseOrder.w
@@ -1,0 +1,558 @@
+&ANALYZE-SUSPEND _VERSION-NUMBER AB_v10r12 GUI
+&ANALYZE-RESUME
+&Scoped-define WINDOW-NAME C-Win
+&ANALYZE-SUSPEND _UIB-CODE-BLOCK _CUSTOM _DEFINITIONS C-Win 
+/*------------------------------------------------------------------------
+
+  File: util/ClosePurchaseOrder.w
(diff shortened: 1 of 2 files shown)


ron-advantzware (changes requested):
ClosePurchaseOrder.w - remove hard coded HELP trigger on the frame and replace with dAOA Lookups.
ClosePurchaseOrderProc.w:
1. transaction scope too large, code should be pessimistic, bf-po-ord and bf-po-ordl should simply be po-ord NO-LOCK and po-ordl NO-LOCK and use a DO TRANSACTION using bf-po-ord and bf-po-ordl respectively to EXCLUSIVE-LOCK and then RELEASE buffer records.  Since this module could be run with a parameter indicating NOT to perform an update, don't want to be waking up the transaction manager for no reason, this could effect performance for no reason.
2. Why using cocode?  Doesn't appear to serve a purpose, simply use ipcCompany and remove any references to cocode.
3. Line 56 has an empty IF ... THEN DO:.  Serves no purpose, remove it.
4. Lines 67 & 68 can be combined into an OR statement.
5. Line 95 & 96 combine into a single ASSIGN.
6. RELEASE lines 113 & 114 are not needed, FOR EACH blocks release their buffers by default.
7. system/OutputProcs.p is instantiated at line 33, need to DELETE PROCEDURE before exiting this program.

PR #9581, opened 2022-02-03
107134 IC3 Count Error

tyndmar:
Programs changed:
fg/r=phce&pN.w - added records-processed limit field
fg/cpostLimit.i - copy of fg-cpost.i with support for records-processed parameter


ron-advantzware (changes requested):
fg/fg-cpostLimit.i - missing lock phrase for
Line 228 oe-ordl (no-lock?)
Line 243 Loadtag (exclusive?)

fg/r-phce&pN.w - missing lock phrase for
Line 1471 fg-bin (no-lock?)
Line 1485 fg-bin (no-lock?)
Line 1519 fg-rctd (??? - depends on the include immediately after table access)

PR #9596, opened 2022-02-14
92814 E-F print tab calculate button new version performance improvement

sakshisinghjkt:
Updated code for EstimateCalcProcs.p for Performance improvement. It being updated to use Dataset/temp-tables instead of DB tables.

Files:
Source/est/EstimateCalcProcsAlt.p
Source/est/dsEstimate.i
Source/est/dsEstimateTT.i
Source/testers/EstimateCalcProcTester.p


brad-advantzware (changes requested):

brad-advantzware (changes requested):
Remove EstimateCalcProcsAlt.p - We can just keep a version of estimateCalcProcs.p in a different propath, or use the feature branch tester vs. develop/integrated test to compare outcomes.  We don't need this file.
Remove dsEstimate.i (just include the ttEstimateCalc.i)

EstimateCalcProcsTester.p should just be cleaned up if we remove EstimateCalcProcsAlt.p

PR #9602, opened 2022-02-16
55317_CEVersion_New_Enable_Vendor_Selection_when_calculating

VarunAgrawalJKT:
Added a new button foe Copy vendor to all estCostMaterial records and fixed the defect to display the correct estimate cost material data when Item type was purchased in the prompt
Files:
Source/est/estCostMaterialList.w
Source/est/EstimateCalcProcs.p


brad-advantzware (changes requested):
PRIVATE pGetAdders(input..., output array).
	Do iCount = 1 to 6...

PUBLIC Estimate_GetAddersList(input..., output Char).
	def output parameter opcAddersList as character.
	def cAdders as extent...

	Run pGetAdders(input, output cAdders
	do iCount = 1 to EXTENT( opcAddersList	

PUBLIC Estimate_GetAddersArray(Input..., output Array).
	Run pGetAdders

PR #9611, opened 2022-02-18
58548_N_K_1_CeVersion_New_Estimate_Board_Cost_for_Combo_Estimate_must…

VarunAgrawalJKT:
…_Calculate_Summarize_for_All_Forms

We have built up a list of like board that will recost the board across all forms on estimate.
New Setting/NK1 - CEAutoRecostBoard have been implemented to automatically recost the board.
Additionally, a "Recost Board" button will be added to the CEPrompt dialog that will allow for this analysis to be done on-demand.

Files:
Source/est/estCostMaterialList.w
Source/est/ RecostBoardEst.p
Source/est/ RecostBoardEst.i


brad-advantzware (commented):
Varun - Don't be discouraged with the amount of feedback I'm giving you here.  This is just to give you an idea of the standards and practices that I would like you to follow.  Overall, you did a good job here.  
One big point of feedback is that I want to avoid copying and pasting code at almost all costs.  We should refactor this procedure and handle the common elements between this and the PO Recost without a pure code copy

brad-advantzware (changes requested):
Varun - looks better. Naming conventions for procedures:
PRIVATE - start with lowercase p (eg. pGetNewCosts)
PUBLIC - start with [Program_name]_Procedure (eg. RecostBoardEst_RecostBoard) in the event this procedure is declared SUPER and doesn't require the handle reference.

PR #9754, opened 2022-04-02
103364 Refactor SendJob and SendJobAMS

rajeshadv:
Files
1. Deployment\SystemData\attribute.json
2. Source\api\SendJobAMS.p
3. Source\api\SendJob.p


ron-advantzware (changes requested):
Missing Record Lock Phrase for tables job-hdr (line 181) and ef (line 186).

PR #9761, opened 2022-04-07
108892_CEVersion_New_Create_Option_CECostSource

VarunAgrawalJKT:
Cost will be calculated as per the cost source.

File changed:
est\EstimateCalcProcs.p
sys\inc\create_nk1.p
ref\sys-ctrl.i


brad-advantzware (changes requested):
Varun - VendorCostThen... Should still look at the vendor cost. table.  It is difficult to tell from this since there are a lot of indentation changes, but it looks like the "GetVendorCost" is only called if it is vendorCostOnly.  This is not correct.  We simply needed to remove the "RETURN".
If the cost is still 0 from the GetVendorCost process, *then* you see if there is a VendorCostThen... option to overwrite the 0 cost.

PR #9768, opened 2022-04-13
108395 Vendor Cost Matrix issue

sewa67:
Files - Source/util/ImportVendCostMtx.p
Source/viewers/dVendCostLevel.w


brad-advantzware (changes requested):
The 99999999.999999 number should be standard among all Vendor Cost procedures.  There is a function inside VendorCostProcs that will return this Max quantity level so that there is *one* place to manage what Max quantity is.  If that number is not large enough, improve it from within VendorCostProcs.p

PR #9788, opened 2022-04-19
101267 Extend SendJob and SendJobAMS to include additional attributes

rajeshadv:
Files
1. Deployment\SystemData\attribute.json
2. api\SendJob.p
3. api\SendJob.i
4. api\SendJobAMS.p

### DIFF
--- a/Deployment/SystemData/attribute.json
+++ b/Deployment/SystemData/attribute.json
@@ -133,6 +133,20 @@
     "rec_key": "202111190040008130282"
   },
   {
+    "dbFieldName": "i-coat",
+    "dbFieldExtent": 0,
+    "attributeName": "Coats",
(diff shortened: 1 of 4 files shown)


ron-advantzware (changes requested):
Procedure pGetJobMaterialInfo (line 205), several Output variables are concatenated then TRIM to remove any trailing commas, however, no comma delimiter is being added to the concatenated variable, and these TRIM to remove commas should be moved and grouped to the last ASSIGN of the procedure.

ron-advantzware (changes requested):
The concatenation of commas needs to be added to the end, which is why we use the TRIM of the commas, putting where you have it will put a leading comma at the start of the variable.

PR #9982, opened 2022-06-03
108422 Performance issue when deleting estimate components

rajeshadv:
Files
1. est\u2kinc2c.p
2. est\kstylec.i
3. est\u2kinc1c.p
4. rfq\u2kinc1c.p
5. rfq\u2kinc2c.p
6. sys\inc\round.i
7. sys\inc\xeb-form.i


ron-advantzware (changes requested):
change to use table buffer on FOR EACH block

PR #10217, opened 2022-09-05
98050 Auto PO from order creation doesn't use customer pricing

sachin448:
File - Source/system/VendorCostSelector.w


ron-advantzware (changes requested):
Cannot manually modify code that is App Builder Generated.  Doing manual changes would be over written if in the future another developer uses App Builder to make modifications.  Please re-do these changes but from within an App Builder session.

